PR #60, opened 2020-01-28
File-system IdentityController for nacl-did

rado0x54:
First implementation of Identitycontroller for nacl-did based on the existing Ethr Identity Controller.
*Do not merge* until the following issue is resolved:

- Signer Interface needs to be generalized finish integration. Dependency on https://github.com/uport-project/daf/issues/51
- See marked //TODOs.


simonas-notcat:
We will probably create an IdentityProvider for `did:key` which is similar to nacl-did

Issue #51, opened 2020-01-09
DID Document management abstraction


simonas-notcat:
- [x] Update `IdentityManager`, `IdentityController` and `Issuer` interfaces
- [x] Update `daf-ethr-did-*` packages to follow new architecture
- [x] Add DID Doc management methods to `daf-cli`
- [ ] Expose these new methods over GraphQL

## Proposed new interfaces

> `sign()`, `encrypt()`, `decrypt()` are placeholders for now

```typescript
export interface Identity {
  identityProviderType: string
  did: string
  didDoc: () => Promise<DIDDocument>
  sign: (data: string, keyId?: string) => Promise<EcdsaSignature | string>
  encrypt: (to: string, data: string | Uint8Array) => Promise<Encrypted>
  decrypt: (encrypted: Encrypted) => Promise<string>
  addPublicKey: (type: string, proofPurpose?: string[]) => Promise<PublicKey>
  removePublicKey: (keyId: string) => Promise<boolean>
  addService: (service: ServiceEndpoint) => Promise<boolean>
  removeService: (service: ServiceEndpoint) => Promise<boolean>
}

export interface IdentityProvider {
  type: string
  description: string
  createIdentity: () => Promise<Identity>
  importIdentity: (secret: string) => Promise<Identity>
  exportIdentity: (did: string) => Promise<string>
  deleteIdentity: (did: string) => Promise<boolean>
  getIdentities: () => Promise<Identity[]>
  getIdentity: (did: string) => Promise<Identity>
  on(event: string | symbol, listener: (...args: any[]) => void): this
  // event: 'identityCreated', 'identityDeleted'
}


export interface IdentityManager {
  getIdentityProviderTypes: () => Promise<string[]>
  createIdentity: (identityProviderType: string) => Promise<Identity>
  importIdentity: (identityProviderType: string, secret: string) => Promise<Identity>
  exportIdentity: (did: string) => Promise<string>
  deleteIdentity: (did: string) => Promise<boolean>
  getIdentities: () => Promise<Identity[]>
  getIdentity: (did: string) => Promise<Identity>
  on(event: string | symbol, listener: (...args: any[]) => void): this
  // event: 'identityCreated', 'identityDeleted'
}
```

PR #61, opened 2020-01-28
Add element fs

christianlundkvist:
Add new Element IdentityController along with new example.


simonas-notcat:
Let's wait until #51 is finished, then we can add necessary changes to this PR and merge it in

simonas-notcat:
Implemented in #145

PR #63, opened 2020-02-10
AsyncStorage identity controller

jasonhealy:
Identity controller using async storage. Useful for POC's in react native.


simonas-notcat:
This will need some changes after #51 is finished

simonas-notcat:
This is solved by https://github.com/uport-project/daf/tree/master/packages/daf-react-native-async-storage

PR #83, opened 2020-02-21
feat: Publishing serviceEndpoint to DID Doc

simonas-notcat:


mirceanis:
this relates to #85

examples/expressjs-ethr/src/identity.ts
```diff
@@ -0,0 +1,39 @@
...
+  return identity
+}
+
+export const setServiceEndpoint = async (identity: Daf.AbstractIdentity, serviceEndpoint: string) => {
+  // Check if DID Document contains current serviceEndpoint
+  const didDoc = await core.didResolver.resolve(identity.did)
+  const exists = didDoc?.service?.find(item => item.serviceEndpoint === serviceEndpoint)
+  if (!exists) {
+    try {
+      const txHash = await identity.identityController.addService({
```
vongohren: @mirceanis @simonas-notcat one question, are you guys baking in tx cost into this, or where are you setting the possibility to cover the cost for this tx?

examples/expressjs-ethr/src/index.ts
```diff
@@ -0,0 +1,224 @@
...
+    } as W3C.ActionSignW3cVc)
+
+    const vpJwt = await core.handleAction({
+      type: W3C.ActionTypes.signVp,
+      did: identity.did,
+      data: {
+        sub: did,
+        vp: {
+          '@context': ['https://www.w3.org/2018/credentials/v1'],
+          type: ['VerifiableCredential'],
```
vongohren: Should this not be a presentation?

examples/expressjs-ethr/src/identity.ts
mirceanis: While there are some defaults, TX cost is ultimately controlled by the web3 provider.

Worth noting that this is an example and not necessarily meant to be used as a general case.

Most likely the general case involves sharing a callback URL along with the the request, such that updating the DID doc is not a requirement.
This is an example that shows did-doc based messaging is possible.

vongohren: @mirceanis yeah I see. Not I just like the easy way of updating the did doc, compared to the necessities today. We want to set a metadata service on a did doc. And cover that cost. But do you have an example of where you setup the web3 provider for that?

mirceanis: Not yet, but feel free to contribute one ;)

examples/expressjs-ethr/src/index.ts
mirceanis: Good catch.

cc @simonas-notcat

examples/expressjs-ethr/src/identity.ts
vongohren: @mirceanis hehe im will try when I dig deeper in this. But do you have any pointers to where you can send in the web3? So I just have start basis :)

examples/expressjs-ethr/src/index.ts
simonas-notcat: @mirceanis we need to fix validation logic in [`did-jwt-vc`](https://github.com/decentralized-identity/did-jwt-vc/blob/develop/src/__tests__/index.ts#L38).

simonas-notcat: @mirceanis I created this PR to fix VP issue: https://github.com/decentralized-identity/did-jwt-vc/pull/20

PR #98, opened 2020-03-11
Add 3id as identity controller

christianlundkvist:
Adding the 3ID DID method as an identity controller.


simonas-notcat:
@christianlundkvist since DAF `2.0.0`, the API to manage identities has changed. This PR is for DAF `1.x` which is not supported anymore

PR #111, opened 2020-03-26
IdentityController: optional ttl and gas with default values

nikolockenvitz:
I would suggest to have `ttl` and `gas` as optional parameters in case someone would like to specify other values for them

### DIFF
--- a/docs/api/daf-core.abstractidentitycontroller.addpublickey.md
+++ b/docs/api/daf-core.abstractidentitycontroller.addpublickey.md
@@ -7,7 +7,7 @@
 <b>Signature:</b>
 
 ```typescript
-addPublicKey(type: string, proofPurpose?: string[]): Promise<string>;
+addPublicKey(type: string, proofPurpose?: string[], ttl?: number, gas?: number): Promise<string>;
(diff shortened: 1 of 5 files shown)


mirceanis:
@nikolockenvitz thanks, this is an interesting proposal.
Of course something like this is necessary to manage an ethr-did but it is harder to justify for the general use-case.
I think there should be some "identity controller params" object exposed by the `AbstractIdentityController` interface that could go into such requests, and the shape of that object would be specific to each identity controller.

So, for an ethr-did it would require Ethereum specific fields like `gasLimit` and `gasPrice`, `RPC`, etc while for some other DID controller it would require a Bearer Token for example, or whatever else that controller needs to do its job.

---

An alternative way to do this that does not require a change to the abstract interface is to have some defaults that a particular IdentityController implementation requires on construction and that could be used throughout the operations that it does.
So, for the ethr-did controller, one could set the desired gas price on application startup or provide a callback that can respond with a dynamic price at the time of call.

@simonas-notcat what do you think?

simonas-notcat:
Fixed in #137

PR #117, opened 2020-04-11
Doc : Error in Example - core should be replaced by agent

devrajsinghrawat:
As core is replaced by agent in DAF 4.0 so if we will use core then code will raise an error

### DIFF
--- a/docs/Docs.md
+++ b/docs/Docs.md
@@ -117,8 +117,8 @@ console.log(`🚀  Server ready at ${info.url}`)
 ## Typescript
 
 ```typescript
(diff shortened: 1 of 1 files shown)


devrajsinghrawat:
Updated the Beta Read doc, as in 4.0 agent is being exposed as core so when we tried to use the given examples as mentioned it was throwing an error so there was update in example code ,, most of the changes proposed are type between core and agent

simonas-notcat:
Fixed in https://github.com/uport-project/daf/commit/568a845e84d710a07bc1077728e3af614fab7c4c

PR #119, opened 2020-04-20
feat: Optionally add permissions to gql resolvers

walfly:
Still need to add more tests.  Allows configuring agent with authentication boolean at instantiation.  An auth-identity can be set in the request context and resolvers will include that identity in their generated queries if authentication is set to true.

### DIFF
--- a/packages/daf-core/src/graphql/graphql-core.ts
+++ b/packages/daf-core/src/graphql/graphql-core.ts
@@ -11,6 +11,8 @@ import {
   Any,
   IsNull,
   FindManyOptions,
+  SelectQueryBuilder,
+  Brackets,
 } from 'typeorm'
 import { Agent } from '../agent'
 import { Message } from '../entities/message'
@@ -21,6 +23,7 @@ import { Identity } from '../entities/identity'
 
 export interface Context {
   agent: Agent
+  authenticatedDid?: string
 }
 
 export interface Order {
@@ -55,19 +58,7 @@ interface TypeOrmOrder {
   [x: string]: 'ASC' | 'DESC'
 }
 
-function transformFindInput(input: FindInput): FindManyOptions {
-  const result: FindManyOptions = {
-    where: {},
-  }
-  if (input?.skip) result['skip'] = input.skip
-  if (input?.take) result['take'] = input.take
-  if (input?.order) {
-    const order: TypeOrmOrder = {}
-    for (const item of input.order) {
-      order[item.column] = item.direction
-    }
-    result['order'] = order
-  }
+function createWhereObject(input: FindInput): any {
   if (input?.where) {
     const where = {}
     for (const item of input.where) {
@@ -113,10 +104,35 @@ function transformFindInput(input: FindInput): FindManyOptions {
       if (item.not === true) {
         where[item.column] = Not(where[item.column])
       }
+      return where
     }
-    result['where'] = where
   }
-  return result
+}
+
+function decorateQB(
+  qb: SelectQueryBuilder<any>,
+  tablename: string,
+  input: FindInput,
+): SelectQueryBuilder<any> {
+  if (input?.skip) qb = qb.skip(input.skip)
+  if (input?.take) qb = qb.take(input.take)
+
+  if (input?.order) {
+    for (const item of input.order) {
+      qb = qb.orderBy(`${qb.connection.driver.escape(tablename)}.${qb.connection.driver.escape(item.column)}`, item.direction)
+    }
+  }
+  return qb
+}
+
+function checkAuthIdentity(did: string, authenticatedDid?: string): boolean {
+  if (!authenticatedDid) {
+    return true
+  }
+  if (authenticatedDid !== did) {
+    throw new Error('can not access restricted properties of another identity')
+  }
+  return true
 }
 
 export interface FindArgs {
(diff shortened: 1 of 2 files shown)


packages/daf-core/src/graphql/graphql-core.ts
```diff
@@ -113,52 +116,190 @@ function transformFindInput(input: FindInput): FindManyOptions {
...
 
 const messages = async (_: any, args: FindArgs, ctx: Context) => {
-  return (await ctx.agent.dbConnection).getRepository(Message).find(transformFindInput(args.input))
+  const where = createWhereObject(args.input)
+  let qb = (await ctx.agent.dbConnection)
+    .getRepository(Message)
+    .createQueryBuilder('message')
+    .where(where)
+  qb = decorateQB(qb, 'message', args.input)
+  if (ctx.agent.authentication) {
```
simonas-notcat, line 151: Since this authentication is impacting only GraphQL resolvers, I would suggest not adding this logic to the `Agent` and keeping it only in GraphQL `Context`. Like this:



``` typescript
export interface Context {
  agent: Agent,
  authenticatedDid: string,
}

// ...

  if (ctx.authenticatedDid) {
    qb = qb.andWhere(
      new Brackets(qb => {
        qb.where('message.to = :ident', { ident: authenticatedDid })
        .orWhere('message.from = :ident', { ident: authenticatedDid })
      }),
    )
  }
```

This way we leave it to application developers to decide how they authenticate requests.

packages/daf-core/src/agent.ts
```diff
@@ -132,4 +138,38 @@ export class Agent extends EventEmitter {
     debug('Handle action %o', action)
     return this.actionHandler.handleAction(action, this)
   }
+
```
simonas-notcat, line 141: I don't think we need to add all these methods just for GraphQL resolvers. This can be solved by passing `authenticatedDid` in GraphQL context. As explained in my other comment

packages/daf-core/src/graphql/graphql-core.ts
```diff
@@ -28,21 +30,34 @@ export interface Order {
...
+//   MoreThanOrEqual = 'MoreThanOrEqual',
+//   | 'Equal'
+//   | 'Like'
+//   | 'Between'
+//   | 'In'
+//   | 'Any'
+//   | 'IsNull'
+// }
+
 export interface Where {
```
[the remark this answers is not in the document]
walfly: It was causing me trouble, i figured it out and set it back

packages/daf-core/src/graphql/graphql-core.ts
walfly: ok, needs to be Promise<string> because context function can't be async

packages/daf-core/src/graphql/graphql-core.ts
```diff
@@ -113,52 +116,190 @@ function transformFindInput(input: FindInput): FindManyOptions {
...
+  qb: SelectQueryBuilder<any>,
+  tablename: string,
+  input: FindInput,
+): SelectQueryBuilder<any> {
+  if (input?.skip) qb = qb.skip(input.skip)
+  if (input?.take) qb = qb.take(input.take)
+
+  if (input?.order) {
+    for (const item of input.order) {
+      qb = qb.orderBy(`"${tablename}"."${item.column}"`, item.direction)
```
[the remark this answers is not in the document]
walfly: hmm, i'm worried that exposes us to some risk of SQL injection

walfly:
@simonas-notcat updated

packages/daf-core/src/graphql/graphql-core.ts
```diff
@@ -113,52 +105,217 @@ function transformFindInput(input: FindInput): FindManyOptions {
...
+  qb: SelectQueryBuilder<any>,
+  tablename: string,
+  input: FindInput,
+): SelectQueryBuilder<any> {
+  if (input?.skip) qb = qb.skip(input.skip)
+  if (input?.take) qb = qb.take(input.take)
+
+  if (input?.order) {
+    for (const item of input.order) {
+      qb = qb.orderBy(`${escape(tablename)}.${escape(item.column)}`, item.direction)
```
simonas-notcat, line 123: No need to use another package for that. You can do the same thing with:
``` typescript
qb = qb.orderBy(`${qb.connection.driver.escape(tablename)}.${qb.connection.driver.escape(item.column)}`, item.direction)
```

packages/daf-core/src/graphql/graphql-core.ts
```diff
@@ -21,6 +24,7 @@ import { Identity } from '../entities/identity'
 
 export interface Context {
   agent: Agent
+  authenticatedDid?: Promise<string>
```
simonas-notcat, line 27: Why is this `Promise<string>` and not just a simple `string`?

I was thinking, that it can be used like that:

``` typescript
const server = new ApolloServer({
  typeDefs: [Daf.Gql.baseTypeDefs, Daf.Gql.Core.typeDefs],
  resolvers: merge(Daf.Gql.Core.resolvers),
  context: async ({ req }) => {
    const authenticatedDid = await applicationSpecificLogic(req.headers.authorization)
    return { agent, authenticatedDid }
  }
})
```

packages/daf-core/src/agent.ts
```diff
@@ -9,6 +9,7 @@ import { ActionHandler } from './action/action-handler'
 import { Action } from './types'
 import { Message, MetaData } from './entities/message'
 import { Connection } from 'typeorm'
+import { verifyJWT } from 'did-jwt'
```
simonas-notcat, line 12: Remove this

packages/daf-core/src/agent.ts
```diff
@@ -34,6 +35,7 @@ interface Config {
 
 export class Agent extends EventEmitter {
   readonly dbConnection: Promise<Connection>
+  readonly authentication?: boolean
```
simonas-notcat, line 38: Remove this

packages/daf-core/src/graphql/graphql-core.ts
```diff
@@ -113,52 +105,217 @@ function transformFindInput(input: FindInput): FindManyOptions {
...
+  if (ctx.authenticatedDid) {
+    const authIdent = await ctx.authenticatedDid
+    qb = qb.andWhere(
+      new Brackets(qb => {
+        qb.where('message.to = :ident', { ident: authIdent }).orWhere('message.from = :ident', {
+          ident: authIdent,
+        })
+      }),
+    )
+  }
```
simonas-notcat, line 162: If we change `authenticatedDid` to `string` then this can be simplified like that:

```typescript
if (ctx.authenticatedDid) {
      qb = qb.andWhere(
       new Brackets(qb => {
          qb.where('message.to = :ident', { ident: ctx.authenticatedDid })
          .orWhere('message.from = :ident', { ident: ctx.authenticatedDid})
       }),
     )
   }
```

simonas-notcat:
@walfly I just released `4.0`. Please create a new PR to `master` with your changes to 

```
packages/daf-core/src/__tests__/resolvers.test.ts
packages/daf-core/src/graphql/graphql-core.ts
```

Sorry for inconvenience

packages/daf-core/src/graphql/graphql-core.ts
walfly: the context function can't be async.  Because verifying a jwt is an async function, this has to be a promise.

packages/daf-core/src/graphql/graphql-core.ts
walfly: https://github.com/apollographql/apollo-server/issues/225

packages/daf-core/src/graphql/graphql-core.ts
walfly: https://github.com/apollographql/apollo-server/issues/225

packages/daf-core/src/graphql/graphql-core.ts
walfly: oh cool, i looked for this, guess I missed it.

packages/daf-core/src/graphql/graphql-core.ts
simonas-notcat: Your mentioned issue is almost 4 years old. I just tried it and it works fine:
https://github.com/uport-project/daf/blob/master/examples/id-hub/src/index.ts#L50

walfly: It was causing linting warning in my ide when I tried before, now its not, must have been something wrong with my config. My mistake, will change.

walfly:
@simonas-notcat alright ready now

Issue #123, opened 2020-04-23
KMS: encrypt private keys


simonas-notcat:
`KeyManagementSystem` should take an encryption key as a constructor param, and then use it to encrypt/decrypt private keys that are stored by the KeyStore

simonas-notcat:
It makes more sense to do this encryption in KeyStore, and not in the KMS

mirceanis:
> It makes more sense to do this encryption in KeyStore, and not in the KMS

:) I was about to comment this when you first posted the issue but see that you already did it in the implementation

PR #128, opened 2020-04-24
SecretBox: encrypt private keys

simonas-notcat:
Fixes: #123 

There is a duplicate implementation of this logic in `daf-libsodium` and `daf-react-native-libsodium`. This can be optimized later


packages/daf-core/src/identity/key-store.ts
```diff
@@ -8,13 +9,19 @@ import Debug from 'debug'
...
     super()
+    if (!secretBox) {
+      console.warn('Please provide SecretBox to the KeyStore')
+    }
   }
 
   async get(kid: string) {
     const key = await (await this.dbConnection).getRepository(Key).findOne(kid)
     if (!key) throw Error('Key not found')
+    if (this.secretBox && key.privateKeyHex) {
```
mirceanis, line 22: I'm still uncomfortable that `Key` objects expose the `privateKeyHex` field.
It can very easily be misused. Even the encrypted form can be used as a private key which is very dangerous since it's not recoverable.

PR #130, opened 2020-04-27
Adding CredentialStatus to data model

simonas-notcat:
Fixes: #129 and adds CredentialStatus to the data model

### DIFF
--- a/packages/daf-cli/src/credential.ts
+++ b/packages/daf-cli/src/credential.ts
@@ -28,6 +28,7 @@ program
         type: 'input',
         name: 'sub',
         message: 'Subject DID',
+        default: identities[0].did
       },
       {
         type: 'input',
@@ -41,6 +42,15 @@ program
         message: 'Claim Value',
         default: 'Alice',
       },
+      {
+        type: 'list',
+        name: 'addStatus',
+        message: 'Add credential status info?',
+        choices: [
+          { name: 'Yes', value: true },
+          { name: 'No', value: false },
+        ],
+      },
     ])
 
     const credentialSubject: any = {}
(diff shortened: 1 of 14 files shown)


packages/daf-cli/src/credential.ts
```diff
@@ -41,22 +42,55 @@ program
         message: 'Claim Value',
         default: 'Alice',
       },
+      {
+        type: 'list',
+        name: 'addStatus',
+        message: 'Add credential status info?',
```
mirceanis, line 48:
```suggestion
        message: 'Is the credential revocable?',
```

packages/daf-w3c/src/action-handler.ts
```diff
@@ -109,54 +114,80 @@ export interface PresentationInput {
...
+        result['vc']['credentialSubject'] = credentialSubject
+      break
+      case '@context':
+      case 'context':
+        result['vc']['@context'] = input[key]
+      break
+      case 'type':
+        result['vc']['type'] = input[key]
+      break
+      case 'expirationDate':
```
mirceanis, line 139: There should also be a case for  `issuanceDate` that translates to `nbf`
Same transformations about date apply to presentations too and should also be done in reverse when parsing JWT

packages/daf-w3c/src/action-handler.ts
```diff
@@ -109,54 +114,80 @@ export interface PresentationInput {
   tag?: string
   expirationDate?: string
   verifiableCredential: string[]
+  [x: string]: any
 }
 
 const transformCredentialInput = (input: CredentialInput): VerifiableCredentialPayload => {
```
awoie, line 120: Are we dropping custom claims outside of the `credentialSubject`, e.g., 

```
{
  "@context": [
    "https://www.w3.org/2018/credentials/v1",
    "https://www.w3.org/2018/credentials/examples/v1"
  ],
  "id": "http://example.edu/credentials/1872",
  "type": ["VerifiableCredential", "AlumniCredential"],
  "issuer": "https://example.edu/issuers/565049",
  "issuanceDate": "2010-01-01T19:73:24Z",
  "customProperty": "12345",
  "credentialSubject": {
    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
    "alumniOf": {
      "id": "did:example:c276e12ec21ebfeb1f712ebc6f1",
    }
  }
```

awoie, line 120: What happens if `credentialSubject` is an array? I don't think we should support this at the moment but the spec supports it, so we should handle that case with an error.

mirceanis: we were before, but no longer after this PR

PR #134, opened 2020-04-30
VP multiple audiences

simonas-notcat:
@walfly I tried cherry-picking your changes into this branch and testing how it works. But I'm having issues making basic query by `audience` field work. Maybe you will have more luck figuring this out


walfly:
@simonas-notcat have it working its kind of a hacky solution. Basically find options don't work very well for many to many relationships, so audience queries get skipped in the initial where-builder and added on separately later

PR #150, opened 2020-05-14
chore: Enforce tighter Typescript and Eslint rulesets

roderik:
By increasing the strictness of the typescript rules, it is possible to catch several series of issues on compile time. The most notable types are:

- implicit any: things that by accident remain types as any, and as such bypass typing
- strict null: makes sure that a value cannot be null. e.g. find in a database returns value or null. You cannot get a field out of the variable until you made sure it is not null/undefined. This catches all "cannot get x from undefined" errors

I also added eslint (configured for Typescript) to provide even better linting checks

I tried to modify as little of the API as possible. Optimizations could be 

- not passing around the Promise of the dbConnection and checking it it resolves only once
- making "raw" a required field on Message


jasonhealy:
Thanks for the PR @roderik - Just a few notes:

- The PR is very big and could be broken down into multiple smaller PR's like `error-handling`, `formatting`, `type-safety`, `optimisations` etc. For type-safety you could focus on one package at a time and then switch over the stricter rule detection once all packages are adhering (rules can be commented out temporarily or target those files specifically). Please provide the instructions on how to test the type safety changes.
- Commit style is conventional so only use `feat` for actual framework features as opposed to development features. Same goes for `fix` - we 'try' to use this for bug fixes from a previous change-set. `build` and `refactor` can be used here.

roderik:
> Thanks for the PR @roderik - Just a few notes:
> 
> * The PR is very big and could be broken down into multiple smaller PR's like `error-handling`, `formatting`, `type-safety`, `optimisations` etc. For type-safety you could focus on one package at a time and then switch over the stricter rule detection once all packages are adhering (rules can be commented out temporarily or target those files specifically). Please provide the instructions on how to test the type safety changes.

changing the tsconfig file, unfortunately, forces you to change it all in one go. You can just close this and use it as a template for when you want to tighten up the typescript.

> * Commit style is conventional so only use `feat` for actual framework features as opposed to development features. Same goes for `fix` - we 'try' to use this for bug fixes from a previous change-set. `build` and `refactor` can be used here.

If you would ever merge this, I would suggest using the suash merge option in github so you can have a single commit in your master changelog.

PR #158, opened 2020-05-18
feat: Expose the CredentialInput type from action-handler

roderik:


mirceanis:
@roderik Can you provide some context about this PR?

The action handler logic will be significantly changed in the next major release and I'd like to understand the issue being resolved here.

PR #167, opened 2020-05-28
Big value handling

ajunge:
Some claims are very big so a bigger field is needed

### DIFF
--- a/packages/daf-core/src/entities/claim.ts
+++ b/packages/daf-core/src/entities/claim.ts
@@ -47,7 +47,7 @@ export class Claim extends BaseEntity {
   @Column()
   type: string
 
-  @Column()
+  @Column("text")
   value: string
 
   @Column()


simonas-notcat:
@mirceanis this is a potential breaking change for anyone with existing databases. Should we mark it as a breaking change and increment the version nr?

ajunge:
> @mirceanis this is a potential breaking change for anyone with existing databases. Should we mark it as a breaking change and increment the version nr?

Why? I think typeorm manage to upgrade the database schema gracefully

mirceanis:
I'm not well versed enough in TypeORM to know if this is breaking or not.
How would one check?
I imagine create a daf-based app, use some claims, upgrade daf to this PR, see if everything still works?

If it doesn't work out of the box, then a migration is necessary and if DAF handles that migration, it doesn't count as a breaking change, because the DB schema is internal to DAF.

PR #176, opened 2020-06-01
fix: return saveDate and updateDate to match the typeDefs

roderik:
The typeDefs define that these fields are there and required

image

But asking for them in the query throws an error

image


roderik:
Never mind, the PR is bad, these types are not on Abstract Identity

PR #178, opened 2020-06-02
Update the keyManagementSystem lib

devrajsinghrawat:
Update the keyManagementSystem lib to daf-react-native-libsodium from daf-libsodium


mirceanis (changes requested):
@devrajsinghrawat thank you for contributing.
Can you provide some context for this PR?
Why is this change required?

PR #207, opened 2020-07-06
Update abstract-identity-provider.ts

ilanolkies:
Minor change


mirceanis:
Good catch. Thank you!

PR #239, opened 2020-09-16
feat: Enable setting headers to DAF DID Comm

ilanolkies:
Allow to set headers to the HTTPS POST request when using DID Comm action handler

### DIFF
--- a/packages/daf-did-comm/src/action-handler.ts
+++ b/packages/daf-did-comm/src/action-handler.ts
@@ -19,7 +19,7 @@ export interface ActionSendDIDComm extends Action {
     type: string
     body: any
   }
-  headers?: HeadersInit
+  headers?: Record<string, string>
 }
 
 export class DIDCommActionHandler extends AbstractActionHandler {
--- a/packages/daf-did-comm/src/action-handler.ts
+++ b/packages/daf-did-comm/src/action-handler.ts
@@ -19,6 +19,7 @@ export interface ActionSendDIDComm extends Action {
     type: string
     body: any
   }
+  headers?: HeadersInit
 }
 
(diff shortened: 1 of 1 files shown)


packages/daf-did-comm/src/action-handler.ts
```diff
@@ -19,6 +19,7 @@ export interface ActionSendDIDComm extends Action {
     type: string
     body: any
   }
+  headers?: HeadersInit
```
mirceanis, line 22: Where is the `HeadersInit` type defined?

mirceanis (changes requested):
This looks good, thank you for starting this.
Can you clarify the question of the data type, please?

packages/daf-did-comm/src/action-handler.ts
ilanolkies: Hi!

https://github.com/node-fetch/node-fetch/blob/master/%40types/index.d.ts#L16

ilanolkies: Is from `node-fetch`. I guess it is a Node.js default type, that's why I didn't import it. Ts build is running ok

mirceanis: I lean more toward declaring it as `headers: Record<string, string>` since that doesn't make it dependant on node-fetch.

What do you think?

ilanolkies: LGTM! Do I have to address the change?

mirceanis:
```suggestion
  headers?: Record<string, string>
```

mirceanis (approved):

Issue #228, opened 2020-08-31
Error `Method importKey not implemented` when calling IdentityManager - > importIdentity(...) on a different device


Muhammad-Altabba:
[Edited]
Hello,

When calling IdentityManager - > exportIdentity and then IdentityManager - > importIdentity(...) on the same machine (having the same sqlite file and running nodejs) the method works. But when calling at on another machine with React Native it throws `Method importKey not implemented`.

Note: this issue emerged when trying to implement Identity export/import according to https://github.com/uport-project/daf/issues/202

I tried to manually import the Key first bust after successfully importing and exporting them, before the Identity import, and then importing the Identities, the same exception came.

I double-checked that the providers are set correctly. Actually, everything else other than this function works well.

However, bellow is the providers' setup code:

```

let KeyManagementSystem, SecretBox;
if (isNode) {  // For NodeJS (the test-cases run within Node)
  // Note: having the following block without 'eval' will cause the react native app to crash!
  //  Unclear why, even when the 'if' condition is false, react-native crash if there is no 'eval'!  
  eval('const libsodium = require("daf-libsodium");\
    KeyManagementSystem = libsodium.KeyManagementSystem;\
    SecretBox = libsodium.SecretBox;');
}
else { // For react-native
  const libsodium = require("daf-react-native-libsodium");
  KeyManagementSystem = libsodium.KeyManagementSystem;
  SecretBox = libsodium.SecretBox;
}

...

const kms = new KeyManagementSystem(keyStore)

...


const identityProviders = [
  new EthrDid.IdentityProvider({
    identityStore: new IdentityStore('rinkeby-ethr', dbConnection),
    kms,
    network: 'rinkeby',
    rpcUrl: 'https://rinkeby.infura.io/v3/' + infuraProjectId,
  }),
]

...


// Initializing the agent
export const agent = new Agent({
  dbConnection,
  didResolver,
  serviceControllers,
  identityProviders,
  actionHandler,
  messageHandler,
})

...

```

DAF version `5.6.5`

What do you suggest?

Thanks,

Muhammad-Altabba:
I guess I figured it out. I will export and import the keys using `/entities/key` before importing the Identities data.

devrajsinghrawat:
Hi @mirceanis .. We are still facing this issue while importing an identity. Could you please provide your insights on the same. 

many thanks

devrajsinghrawat:
Hi Team,

We have further checked this issue and found the below 

[...]

PR #257, opened 2020-11-09
Import DID support for daf-react-native-libsodium

devrajsinghrawat:
#228 

Feature branch to support **import Identity** for daf-react-native-libsodium package

### DIFF
--- a/packages/daf-react-native-libsodium/src/key-management-system.ts
+++ b/packages/daf-react-native-libsodium/src/key-management-system.ts
@@ -98,4 +98,9 @@ export class KeyManagementSystem extends AbstractKeyManagementSystem {
     debug('Deleting', kid)
     return await this.keyStore.delete(kid)
   }
+
+  async importKey(serializedKey: SerializedKey) {
+    await this.keyStore.set(serializedKey.kid, serializedKey)
+    return new Key(serializedKey)
+  }
 }


packages/daf-react-native-libsodium/src/key-management-system.ts
```diff
@@ -98,4 +98,9 @@ export class KeyManagementSystem extends AbstractKeyManagementSystem {
     debug('Deleting', kid)
     return await this.keyStore.delete(kid)
   }
+
+  async importKey(serializedKey: SerializedKey) {
+    await this.keyStore.set(serializedKey.kid, serializedKey)
```
mirceanis, line 103: There should be some checks regarding the key capabilities.
Either here, or maybe at the `keyStore` level.
For example, if a `SerializedKey` is provided without a `privateKeyHex` component then it will be unusable after import.
Depending on the `keyStore` or the type of key other checks may be needed.

mirceanis (commented):
Thank you for the initiative to contribute.
I'll mark this as a work in progress. Please signal when it's ready for review.

We are getting closer to releasing version 7 which has some significant changes that may lead to conflicts. For this reason, it might be better if you base your changes off of the `beta` branch. That way, they will be easier to incorporate into the codebase.

packages/daf-react-native-libsodium/src/key-management-system.ts
devrajsinghrawat: Hi @mirceanis ,

This fix is for **daf-react-native-libsodium version 6** where the corresponding **daf-libsodium version 6** has the import function implemented in the exact fashion without any checks.

A code changes corresponds to version 7 can have a different PR.

Thanks

mirceanis: ok, so this is not work in progress actually. It makes sense as a fix to daf v6.

I will be merging a bunch of PRs and it will be included in a next hotfix release for daf

mirceanis (approved):

PR #262, opened 2020-11-16
feat: Add event system to agent

mirceanis:
# DAF event system

The event system provides a way for components to interact in a very decoupled way.
DAF agents provide a very simple event system where events can be emitted by anyone and consumed asynchronously by event listeners.

## Emitting an event

The basic call is:
```typescript
agent.emit(type, data)
```

Normally, events follow the fire-and-forget pattern. The emitter should not care how or if the event is interpreted.

However, there may be situations where an application may need to make sure all events are consumed before exiting, or that some state change has been performed by an event listener before going forward.

For this situation, there is another pattern that can be used:
```typescript
await agent.emit(type, data)
```

## Listening for an event.

Listeners are registered at the time of agent creation, and are declared alongside the plugin array. In fact, agent plugins can also behave as event listeners.
A listener must declare `eventTypes` that it can handle and an `async onEvent({type, data}, context)` method that will be called when an event is fired.

```typescript
const fooEventLogger: IEventListener {
  eventTypes: ['fooEvent']
  onEvent: async (event, context) => { console.log(event.data) }
}

const agent = new Agent({
  plugins: [fooEventLogger,...],
})
```

## Listening for multiple event types

Event listeners can register for multiple event types by using multiple entries in the `eventTypes` array.
The same `onEvent` method will be called for all types, so it is up to the listener implementation to differentiate between the events if necessary.

```typescript
const foobarPlugin: IEventListener = {
  eventTypes: ['fooEvent', 'barEvent'],
  onEvent: async (event, context) => {
    if (event.type === 'fooEvent') {
      const fooData = event.data as FooData
      //do something with fooData
    } else if (event.type === 'barEvent') {
      const barData = event.data as BarData
      //do something with barData
    }
  }
}
```

## Using `context`

The `onEvent` method gets access to the `context` object used throughout the agent API.
Using this object, it can call other methods on the agent( for example, calling some plugin method ) or even emit other events.

```typescript
const myEventListener: IEventListener {
  eventTypes: ['myEvent']
  onEvent: async ({type, data}, context) => {
    //call agent methods
    context.agent.someOtherPluginMethod(...);
    //emit other events
    context.agent.emit('otherEvent', otherEventData)
  }
}
```

## Error handling

In case an `Error` is thrown during the processing of an event, the error is emitted as a new event of type `"error"` with the `Error` instance as the event data.
Handling errors, therefore means registering an error listener on the agent.

```typescript
const faultyListener: IEventListener {
  eventTypes: ['fooEvent']
  onEvent: async () => { throw new Error('crashing!!'); }
}

const errorLogger: IEventListener {
  eventTypes: ['error']
  onEvent: console.error
}

const agent = new Agent({
  plugins: [faultyListener, errorLogger],
})
```

---

## Notes

Feedback is very welcome about the use and API of this event system.
Some hotspots for feedback may be the requirement of registering listeners in advance, and maybe the asynchronous treatment of event handlers.

### tasks
* [X] add `IEventListener` interface
* [X] make `IAgentPlugin` an `IEventListener`
* [X] add `emit()` method to `IAgent`
* [X] make `IAgentPlugin.methods` optional; enables listener-only plugins
* [X] call listeners when events are emitted
* [X] remove `EventEmitter` implementation from `MessageHandler`; use `context.agent.emit()` instead
* [x] add tests
* [x] figure out synchronicity. the underlying `EventEmitter` is synchronous.
* [x] prevent endless loops in `error` listener

### not in scope
* predicate logic for filtering events. Events are filtered using string matching on eventType
* registering a listener for ALL events or for glob syntax matching is also not possible in this release
* event intercepting. In a future release, there may be ways of intercepting an event and stopping further propagation.
* type safety of event data. At the moment it is up to developers to use the proper data types wen emitting and listening for events of a particular type.


simonas-notcat (approved):
Looks good

jasonhealy:
Looks good. 

The use case in my mind that if this could solve would be awesome. I’d like to be able to write a configurable walletconnect plugin for daf. Currently that custom event system is spread across a bunch of places in daf mobile. It seems like this will work as most of the event stuff is there just to communicate with daf core and emit events based on messages. 

https://github.com/uport-project/daf-mobile/blob/feature/daf-7-upgrade/src/providers/WalletConnect.tsx

https://github.com/uport-project/daf-mobile/blob/feature/daf-7-upgrade/src/components/WalletConnect.tsx

jasonhealy (approved):

mirceanis:
It looks like the daf event system could be used instead of `wcEventHub` from your example, but I have no idea if it is worth trying to replace things like this.
The event system that I'm proposing here is not meant to be the single or main backbone of an app, but it should serve as an internal backbone for inter-component chatter; all while allowing the app that instantiates it to hook into that same channel to pipe events into it and to listen for specific events coming from plugins.

Of course, it could all be built on top of a fully exposed `EventEmitter` but that would prevent some of the things I'd like to achieve here:
* context passing
* fire-and-forget events
* error collecting
* [future] manifest based permissions
  * plugins would only use events that they are permitted to use based on a pre-declared manifest
  * specific method calls and event types would have a security context associated for a really granular authorization mechanism.

PR #282, opened 2020-11-25
Web DID path (multi user) support 

simonas-notcat:
# Setup

```typescript
import express from 'express'
import { agent } from './agent'
import { AgentRouter, ApiSchemaRouter, DidDocRouter } from 'daf-express'

const getAgentForRequest = async (req: express.Request) => agent
const exposedMethods = agent.availableMethods()
const basePath = '/agent'
const schemaPath = '/open-api.json'

const agentRouter = AgentRouter({
  getAgentForRequest,
  exposedMethods,
})

const schemaRouter = ApiSchemaRouter({
  basePath,
  getAgentForRequest,
  exposedMethods,
})

const didDocRouter = DidDocRouter({
  getAgentForRequest
})

const app = express()
app.use(basePath, agentRouter)
app.use(schemaPath, schemaRouter)
app.use(didDocRouter)
app.listen(3002)
```

# Usage

```typescript
// 'did:web:example.com'
const serverIdentity = await agent.identityManagerCreateIdentity({
  alias: 'example.com',
  provider: 'did:web',
})

// 'did:web:example.com:alice'
const alice = await agent.identityManagerCreateIdentity({
  alias: 'example.com:alice',
  provider: 'did:web',
})
```

### DIFF
--- a/packages/daf-cli/src/server.ts
+++ b/packages/daf-cli/src/server.ts
@@ -85,8 +85,6 @@ program
      * Exposing OpenAPI schema
      */
     console.log('🗺  OpenAPI schema', baseUrl + options.schemaPath)
-    console.log('🧩 Available methods', agent.availableMethods().length)
-    console.log('🛠  Exposed methods', exposedMethods.length)
     app.use(
       options.schemaPath,
       ApiSchemaRouter({
@@ -96,13 +94,6 @@ program
         securityScheme,
       }),
     )
-
-    /**
-     * Handling 'did:web' requests ('/.well-known/did.json' and '/^\/(.+)\/did.json$/')
-     * warning: 'did:web' method requires HTTPS (that is one of the reasons to use ngrok for development)
-     */
-    console.log('📋 DID Document ' + baseUrl + didDocEndpoint)
-    app.use(WebDidDocRouter({ getAgentForRequest }))
 
     console.log('📖 API Documentation', baseUrl + options.apiDocsPath)
     app.use(
@@ -114,6 +105,10 @@ program
         },
       }),
     )
+
+    console.log('🧩 Available methods', agent.availableMethods().length)
+    console.log('🛠  Exposed methods', exposedMethods.length)
+
 
     /**
      * Creating server identity and configuring messaging service endpoint
@@ -156,6 +151,14 @@ program
     }
 
     /**
+     * Handling 'did:web' requests ('/.well-known/did.json' and '/^\/(.+)\/did.json$/')
+     * warning: 'did:web' method requires HTTPS (that is one of the reasons to use ngrok for development)
+     */
+    console.log('📋 DID Document ' + baseUrl + didDocEndpoint)
+    console.log('📋 DID Documents ' + baseUrl + '/(.+)/did.json')
+    app.use(WebDidDocRouter({ getAgentForRequest }))
+
+    /**
      * Serving homepage
      */
     app.get('/', async (req, res) => {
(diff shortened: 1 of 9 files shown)


packages/daf-express/src/did-doc-router.ts
```diff
@@ -0,0 +1,89 @@
...
+   */
+  getAgentForRequest: (req: Request) => Promise<TAgent<IIdentityManager>>
+}
+/**
+ * Creates a router that serves `did:web` DID Documents
+ *
+ * @param options - Initialization option
+ * @returns Expressjs router
+ */
+export const DidDocRouter = (options: DidDocRouterOptions): Router => {
```
mirceanis, line 30: Perhaps we should be more explicit about serving a `did:web` here by naming this `WebDidDocRouter`

packages/daf-cli/src/server.ts
```diff
@@ -27,150 +27,162 @@ program
...
-    const agentRouter = AgentRouter({
-      basePath: apiBasePath,
-      getAgentForRequest: async (req) => agent,
-      exposedMethods,
-      serveSchema: false,
-    })
+    /**
+     * Handling 'did:web' requests ('/.well-known/did.json' and '/^\/(.+)\/did.json$/')
+     */
+    console.log('📋 DID Document ' + baseUrl + didDocEndpoint)
```
mirceanis, line 103: `did:web` won't work unless it is served under a `https` url.
Here may be a good spot to warn users about this aspect.

packages/daf-cli/src/server.ts
```diff
@@ -27,150 +27,162 @@ program
...
+     * Serving homepage
+     */
+    app.get('/', async (req, res) => {
+      const links = [
+        { label: 'API Docs', url: options.apiDocsPath },
+        { label: 'API Schema', url: options.schemaPath },
+        { label: 'DID Document', url: didDocEndpoint },
+      ]
+
+      let verifiablePresentation
```
mirceanis, line 167: I don't think it's a good idea to list all the presentations without auth.
This is great when getting started, and for debugging but it should be opt-in, at least.

mirceanis (changes requested):
It looks good.
I have suggested some minor improvements and posted a comment about listing presentations on the homepage.
Please take a look

packages/daf-express/api/daf-express.api.md
```diff
@@ -5,18 +5,40 @@
 ```ts
 
 import { IAgent } from 'daf-core';
+import { IIdentityManager } from 'daf-core';
 import { Request as Request_2 } from 'express';
```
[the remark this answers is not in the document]
simonas-notcat: This is because of the weird `api-extractor` behaviour

packages/daf-express/src/did-doc-router.ts
simonas-notcat: Makes sense. Renamed to `WebDidDocRouter`

packages/daf-cli/src/server.ts
simonas-notcat: Added comment about this

packages/daf-cli/src/server.ts
simonas-notcat: Added explanation of how this works. This is experimental

mirceanis: 👍 saw it.
I still think this should be opt-in, though?

mirceanis (approved):
looks good

PR #293, opened 2020-12-09
fix(daf-did-jwt): Fix parsing of JWT with missing `typ` in header

### LINKED ISSUE #291: daf7 with schemaValidation on can't use universal resolver
**Bug severity**
3

**Describe the bug**
When using `DafResolverUniversal`, the DID documents are formatted closer to did core spec (they have @context as array of strings)
This causes failures with INVALID_TYPE errors when resolving DIDs

**To Reproduce**
```typescript
import { DafUniversalResolver } from 'daf-resolver-universal'
import { createAgent, IResolver } from 'daf-core'
;(async ()=>{
  const agent = createAgent<IResolver>({
    schemaValidation: true,
    plugins: [
      new DafUniversalResolver({ url: 'https://dev.uniresolver.io/1.0/identifiers/'}),
      //...
    ],
  })
    const doc = await agent.resolveDid({ didUrl: 'did:key:z6Mkfriq1MqLBoPWecGoDLjguo1sB9brj6wT3qZ5BxkKpuP6' })
})();
```

**Expected behaviour**
A `doc` is returned

**Observed behaviour**
`INVALID_TYPE` error is thrown

**Additional context**
The data types in question are imported from `did-resolver`, sometimes indirectly through `did-jwt`.
The latest versions of both these libraries have fixed these definitions.

 - Daf: 7.0.0-beta.55

mirceanis:
`JwtMessageHandler` can parse them but they fail later in the the message handler chain if `typ` is missing.

Also, this fixes #291

### DIFF
--- a/packages/daf-did-jwt/src/message-handler.ts
+++ b/packages/daf-did-jwt/src/message-handler.ts
@@ -19,7 +19,7 @@ export class JwtMessageHandler extends AbstractMessageHandler {
         const resolver = { resolve: (didUrl: string) => context.agent.resolveDid({ didUrl }) }
         const verified = await verifyJWT(message.raw, { resolver, audience })
         debug('Message.raw is a valid JWT')
-        message.addMetaData({ type: decoded.header.typ, value: decoded.header.alg })
+        message.addMetaData({ type: decoded.header.typ || 'JWT', value: decoded.header.alg })
         message.data = verified.payload
       } catch (e) {
         debug(e.message)
--- a/packages/daf-core/package.json
+++ b/packages/daf-core/package.json
@@ -25,7 +25,7 @@
   },
   "devDependencies": {
     "@types/debug": "^4.1.5",
-    "did-resolver": "2.1.1",
+    "did-resolver": "2.1.2",
     "typescript": "^4.0.3"
   },
   "files": [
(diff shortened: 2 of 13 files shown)


packages/daf-did-jwt/src/message-handler.ts
```diff
@@ -19,7 +19,7 @@ export class JwtMessageHandler extends AbstractMessageHandler {
         const resolver = { resolve: (didUrl: string) => context.agent.resolveDid({ didUrl }) }
         const verified = await verifyJWT(message.raw, { resolver, audience })
         debug('Message.raw is a valid JWT')
-        message.addMetaData({ type: decoded.header.typ, value: decoded.header.alg })
+        message.addMetaData({ type: decoded.header.typ || 'JWT', value: decoded.header.alg })
```
simonas-notcat, line 22: Do we need to make any changes here? https://github.com/uport-project/daf/blob/beta/packages/daf-w3c/src/message-handler.ts#L51

simonas-notcat (commented):
Looks good. But not sure about `daf-w3c` message handler

packages/daf-did-jwt/src/message-handler.ts
mirceanis: @simonas-notcat Yes, definitely, but it's not in the scope of this PR

daf-w3c benefits from this PR, though

PR #295, opened 2020-12-10
fix: JSON schema/swagger codegen workaround for VerifiableCredential

tobek:
Not sure if we want to do this, just putting it up for discussion, see https://consensysmesh.slack.com/archives/C01AJDJDRJ7/p1607627004057100?thread_ts=1607397897.054800&cid=C01AJDJDRJ7.

We would have to do the same thing for `VerifiableCredential.proof` property - it only specifies `type` and so codegen library ignores e.g. the `jwt` property unless we a) explicitly include it in schema or b) make `proof` a generic object like this PR does with `credentialSubject`.

### DIFF
--- a/packages/daf-cli/src/schema.ts
+++ b/packages/daf-cli/src/schema.ts
@@ -35,10 +35,20 @@ function createSchema(generator: TJS.SchemaGenerator, symbol: string) {
   let fixedSymbol = symbol.replace('Array<', '').replace('>', '')
 
   const schema = generator.createSchema(fixedSymbol)
-  
-  if (fixedSymbol === 'ICreateVerifiableCredentialArgs') {
+
+  if (schema.definitions?.['VerifiableCredential']) {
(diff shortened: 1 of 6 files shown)


mirceanis:
We need to re-decide what to do about this after we re-investigate some schema generation issues.
cc @simonas-notcat

tobek:
I ended up giving up for now on using the codegen PHP library due to https://github.com/swagger-api/swagger-codegen/issues/8108 and other issues mentioned in my slack thread above. Maybe there are elegant workarounds but the only workarounds I could think of were pretty hacky. If there were a strong incentive to make it work we could make it work, but using the REST API for now is perfectly passable.

Your call if you want to close this PR or leave it open for reference and possible future implementation.

mirceanis:
let's close this since there seems to be no urgency and we can bring it back up once it becomes an issue again

PR #309, opened 2020-12-17
feat: Simpler create VC/VP

simonas-notcat:
This was branched of rename branch. Only the last commit matters here


mirceanis (approved):
Looks good

PR #350, opened 2021-01-26
feat(core): ability to define agent context type

simonas-notcat:


packages/core/src/agent.ts
```diff
@@ -309,7 +309,9 @@ export class Agent implements IAgent {
  * @returns configured agent
  * @public
  */
-export function createAgent<T extends IPluginMethodMap>(options: IAgentOptions): TAgent<T> {
+export function createAgent<T extends IPluginMethodMap, C = Record<string, any>>(
```
mirceanis, line 312: This looks like a breaking change, unless we can make the context type optional.
How does typescript behave when the dev doesn't supply the context type?

simonas-notcat: It is optional. Added a unit test showing that it is optional

mirceanis (approved):
looks good

Issue #362, opened 2021-02-02
[proposal] Support URL redirects in url-handler


rkreutzer:
Some applications use a URL shortener service to create QR Codes that are easier to scan. When fetching that URL, a 302 redirect returns the full URL that includes a JWT. 

The url-handler could be revised to compare the URL that was returned in the fetch command with the one submitted. If they are different, it could check if a JWT exists on the c_i query parameter of the returned URL.

mirceanis:
Sounds good.
Do you want to try to implement this functionality, @rkreutzer ?

rkreutzer:
I can do that.

mirceanis:
Your PR to fix this is now merged.
It will be available in the next minor release.

In the meantime, you can use the `@next` release channel: `@veramo/url-handler@next`
Generally, it's best to use the same release with all your @veramo dependencies.

PR #366, opened 2021-02-03
feat(url-handler): allow for URL redirects (#362)

rkreutzer:
I couldn't find a way for fetch mocks to return the redirected URL, so I didn't add a test.


mirceanis (approved):
Looks great!

I added a couple of tests that exercise the new code and removed some un-necessary `package-lock.json` files.

Thank you for your contribution!

PR #384, opened 2021-02-15
fix: use URI encoded host in web-did-doc-router

### LINKED ISSUE #383: remote server does not support web did with port
**Bug severity**
1

**Describe the bug**
According to https://github.com/w3c-ccg/did-method-web/issues/7 to use web did with port, the host should be URI encoded

**To Reproduce**
1. start remote server on port 3000
2. generate an identifier for localhost:3000
2. navigate to localhost:3000/.well-known/did.json

**Observed behaviour**
An empty json document is returned instead of a valid document
   
**Expected behaviour**
A DID document for localhost:3000 should be returned

 - Veramo: 1.1.0
 - Browser: any
 - Node Version: 12.20.0

sk91:
Closes #383

### DIFF
--- a/packages/remote-server/src/web-did-doc-router.ts
+++ b/packages/remote-server/src/web-did-doc-router.ts
@@ -55,11 +55,15 @@ export const WebDidDocRouter = (options: WebDidDocRouterOptions): Router => {
     return didDoc
   }
 
+  const getAliasForRequest = async (req: Request) => {
+    return encodeURIComponent(req.get('host') || req.hostname)
+  }
+
   router.get(didDocEndpoint, async (req: RequestWithAgentDIDManager, res) => {
     if (req.agent) {
       try {
         const serverIdentifier = await req.agent.didManagerGet({
-          did: 'did:web:' + req.hostname,
+          did: 'did:web:' + getAliasForRequest(req),
         })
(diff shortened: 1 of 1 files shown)


mirceanis:
Thank you for the contribution.

This depends on how the `did:web` spec will evolve, after which the [web-did-resolver](https://github.com/decentralized-identity/web-did-resolver) will also have to be upgraded to match the spec.
It is very likely that the proposal won't pass, because the `%` character is not part of the `method-specific-id` portion of the [DID syntax](https://www.w3.org/TR/did-core/#did-syntax).
I will mark this as `blocked` until the web-did spec and the resolver are upgraded.

Also, keep in mind that DIDs such as `did:web:localhost%3A8080` will still need to be resolved using `https`, so you will need a valid certificate for localhost

sk91:
Yeh, it seems like I was to quick to act here :)

mirceanis:
> Yeh, it seems like I was to quick to act here :)

I didn't mean to put a wrench in your machine :)
Your contribution is very much appreciated and I look forward to your next proposals!

We are fully aware that all the specs we're working against are in flux, so one of the goals of this framework is to be an enabler for quick experimentation, like the proposal you made here.

mirceanis (approved):
This was recently unblocked so it makes sense to merge it now.
Thank you for your contribution!

PR #413, opened 2021-03-09
fix(cli): validate config file version number

simonas-notcat:


packages/cli/src/server.ts
```diff
@@ -24,6 +24,15 @@ program
   .description('Launch OpenAPI server')
   .option('-p, --port <number>', 'Optionally set port to override config')
   .action(async (cmd) => {
+    const { version } = createObjects(getConfig(program.opts().config), {
+      version: '/version',
+    })
+
+    if (version !== 1) {
```
mirceanis, line 31: I think it should also check if a version is specified:
```suggestion
    if (version && version !== 1) {
```

and perhaps be lenient with the type with `!=` instead of `!==`?

mirceanis (changes requested):
good idea!
I believe this check is not only specific to server config. It should probably be done in `createObjects()` or `getConfig()` directly.
What do you think?

PR #447, opened 2021-04-08
feat: add MemoryDIDStore and MemoryKeyStore

simonas-notcat:
In memory DIDStore for DIDManager and KeyStore for KeyManager. Usefull in situations when DIDs and Keys don't have to be stored in DB and can be imported after agent initialization. For example when using MetaMask as KMS


mirceanis (approved):
looks good

PR #457, opened 2021-04-16
ci: add @experimental release channel

mirceanis:
This adds an experimental channel for releases that are too early to mint into a stable API


simonas-notcat (approved):

package.json
```diff
@@ -61,15 +62,6 @@
   "publishConfig": {
     "access": "public"
   },
-  "release": {
```
rado0x54, line 64: This is legacy I assume?

rado0x54 (approved):

package.json
mirceanis: yup

PR #480, opened 2021-04-26
add names to entities (so that code on devices succeeds)

trentlarson:
After deploying in the App & Play Stores, I got an error with the TypeORM code working with these entities.  This fixed the problem.


trentlarson:
FYI I realize this PR is lacking in detail, but now that I've resolved the issue I cannot find the original error that showed on the platforms.  The tough thing was that all my entity SQL was failing and I got no error toast message so I had to do my own debugging and I guess I didn't save the error messages anywhere.  So I'm hoping that this is safe enough to accept, since it simply makes the entity names explicit.  If you need me to go back and release a bad version of the app and try to catch the error then... I might work on that, but it'll take a while.

simonas-notcat (approved):
Looks good!

PR #494, opened 2021-04-26
fix(ci): explicitly publish to `@latest` from `main` branch

mirceanis:
This is an attempt to fix `lerna ERR! E400 Must associate a single dist tag with new publishes.` that [we got on `main`](https://github.com/uport-project/veramo/runs/2438794873?check_suite_focus=true#step:14:126).


jasonhealy (approved):
Looks good

PR #503, opened 2021-05-03
fix(cli): did:key dependency

simonas-notcat:


mirceanis (approved):
looks good.
merging this should release version 1.2.1 with `@latest` dist-tag

Issue #468, opened 2021-04-20
Define KMS API


awoie:
1. KMS API should support low-level crypto API, also including those required for ECDH-1PU:
- sign
- sharedSecret(myPrivKeyHandle, publicKeyRecipient, 'X25519')
- ... 
2. Each `AbstractKeyManagementSystem` implementation should be in charge of storing the private key material, instead of relying on `@veramo/key-manager` for storage. Private keys should not be exposed between plugins.

pre-work here: https://whimsical.com/crypto-suites-kms-link-BycG8ABUoV3L4MEGUkHTfG

mirceanis:
TODO:
* [ ] remove privateKey references from `IKey` interface. key material should not leak outside the KMS that manages it
* [ ] redefine low level crypto methods that can be provided (`sign`, `decrypt`, `sharedSecret`, please comment if you can think of another one)
* [ ] KMS layer should report which plugins it supports, and add this metadata to the `IKey` during create/import.
* [ ] create a localStorage plugin interface that can be used by the new KeyManager ( and possibly DIDManager ) to persist the links between kid and KMS implementations
* [ ] explore the decoupling of the database storage from key storage (perhaps using an encryption layer on top of the localStorage plugin)
* [ ] ensure easy migration path from `@veramo/key-manager` with `dbConnection` to the new system.

mirceanis:
I'm splitting this up into parts so that other work is not blocked for the whole duration of this implementation.
1. expose the `sign`, `decrypt`, `sharedSecret` methods on the current key-manager to unblock other tasks
2. continue working on the refactoring
3. adapt other plugins to the potentially new API after refactoring.

mirceanis:
Closed by #529

PR #529, opened 2021-05-19
feat(key-manager): add generic signing capabilities

mirceanis:
This PR extends the signing capabilities of the `key-manager` plugin.

The `key-manager` plugin now exposes a `keyManagerSign` method that accepts extra `alg`, `enc` parameters and supports extension of the parameter object.
The `enc` parameter indicates the encoding of the `data` string, defaulting to `utf-8` if absent.
The `alg` option should indicate an algorithm to be used.
Depending on the capabilities of the KMS (one of the `AbstractKeyManagementSystem` implementations), different signatures can be requested.

For example:
* a KMS implementation that has direct access to key material for a `Secp256k1` key should be able to support:
  * `ES256K`, `ES256K-R`, `eth_signTransaction`, `eth_signTypedData`, `eth_signMessage`
* a KMS implementation that is based on a `web3` abstraction, would only be able to do:
  *  `eth_signTransaction`, `eth_signTypedData`, and `eth_signMessage`

The capabilities of the KMS are recorded in the `IKey.meta.algorithms` array during key creation or import.
Further improvements can be made to programmatically filter for KMS or key capabilities.

This PR relates to #468

### Examples:
Sign a JWS payload with ES256K:
```typescript
// header with `"alg": "ES256K"`
const jwsBundle = `${encodedHeader}.${encodedPayload}`
// for this alg,the result is base64url encoded
const signature = await agent.keyManagerSign({
  algorithm: 'EdDSA',
  data: jwsBundle,
  encoding: 'utf-8', // this is the default encoding, so it could be omitted in this case
  keyRef: key.kid
})
const jwsString = `${encodedHeader}..${signature}`
```

Sign an ETH transaction:
```typescript
// RLP encode the transaction fields
const txData = serialize({
  to: '0xce31a19193d4b23f4e9d6163d7247243bAF801c3',
  value: 300000,
  gasLimit: 43092000,
  gasPrice: 20000000000,
  nonce: 1,
})

// compute the raw transaction string
// for transaction signing, the output is the `0x` prefixed hex string representing the signed raw transaction.
const rawTx = await agent.keyManagerSign({
  algorithm: 'eth_signTransaction',
  data: txData,
  encoding: 'hex',
  keyRef: key.kid
})
// rawTx = '0xf869018504a817c80084029.......'
```

### DIFF
--- a/packages/core/src/types/IIdentifier.ts
+++ b/packages/core/src/types/IIdentifier.ts
@@ -71,9 +71,14 @@ export interface IKey {
   privateKeyHex?: string
 
   /**
-   * Optional. Key metadata. Can be used to store auth data to access remote kms
+   * Optional. Key metadata. This should be used to determine which algorithms are supported.
    */
-  meta?: object | null
+  meta?: KeyMetadata | null
+}
+
+export interface KeyMetadata {
+  algorithms?: string[]
+  [x: string]: any
 }
 
 /**
--- a/packages/core/src/types/IKeyManager.ts
+++ b/packages/core/src/types/IKeyManager.ts
@@ -87,15 +87,17 @@ export interface IKeyManagerDecryptJWEArgs {
  */
 export interface IKeyManagerSignArgs {
   /**
-   * Key ID
+   * The key handle, as returned during `keyManagerCreateKey`
    */
-  kid: string
+  keyRef: string
 
   /**
    * The algorithm to use for signing.
    * This must be one of the algorithms supported by the KMS for this key type.
+   * 
+   * The algorithm used here should match one of the names listed in `IKey.meta.algorithms`
    */
-  alg?: string
+  algorithm?: string
 
   /**
    * Data to sign
@@ -105,7 +107,7 @@ export interface IKeyManagerSignArgs {
   /**
    * If the data is a "string" then you can specify which encoding is used. Default is "utf-8"
    */
-  enc?: 'utf-8' | 'base16' | 'base64' | 'hex'
+  encoding?: 'utf-8' | 'base16' | 'base64' | 'hex'
 
   [x: string]: any
 }
--- a/packages/core/src/types/IKeyManager.ts
+++ b/packages/core/src/types/IKeyManager.ts
@@ -1,5 +1,5 @@
 import { IPluginMethodMap } from './IAgent'
-import { TKeyType, IKey } from './IIdentifier'
+import { TKeyType, IKey, KeyMetadata } from './IIdentifier'
 
 /**
  * Input arguments for {@link IKeyManager.keyManagerCreate | keyManagerCreate}
@@ -19,7 +19,7 @@ export interface IKeyManagerCreateArgs {
   /**
    * Optional. Key meta data
    */
-  meta?: object
+  meta?: KeyMetadata
 }
 
 /**
@@ -79,6 +79,35 @@ export interface IKeyManagerDecryptJWEArgs {
    * Encrypted data
    */
   data: string
+}
+
+/**
+ * Input arguments for {@link IKeyManager.keyManagerSign | keyManagerSign}
+ * @public
+ */
+export interface IKeyManagerSignArgs {
+  /**
+   * Key ID
+   */
+  kid: string
+
+  /**
+   * The algorithm to use for signing.
+   * This must be one of the algorithms supported by the KMS for this key type.
+   */
+  alg?: string
+
+  /**
+   * Data to sign
+   */
+  data: string
+
+  /**
+   * If the data is a "string" then you can specify which encoding is used. Default is "utf-8"
+   */
+  enc?: 'utf-8' | 'base16' | 'base64'
+
+  [x: string]: any
 }
 
 /**
@@ -142,6 +171,13 @@ export interface IKeyManager extends IPluginMethodMap {
    * Imports a created key
    */
   keyManagerImport(args: IKey): Promise<boolean>
+
+  /**
+   * Generates a signature according to the algorithm specified.
+   * @throws `Error("not_supported")` if the KMS does not support the operation or if the key does not match the algorithm.
+   * @param args
+   */
+  keyManagerSign(args: IKeyManagerSignArgs): Promise<string>
 
   /**
    * Encrypts data
--- a/packages/core/src/types/IKeyManager.ts
+++ b/packages/core/src/types/IKeyManager.ts
@@ -105,7 +105,7 @@ export interface IKeyManagerSignArgs {
   /**
    * If the data is a "string" then you can specify which encoding is used. Default is "utf-8"
    */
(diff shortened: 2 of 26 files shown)


packages/core/src/types/IKeyManager.ts
```diff
@@ -81,6 +81,35 @@ export interface IKeyManagerDecryptJWEArgs {
...
+
+  /**
+   * Data to sign
+   */
+  data: string
+
+  /**
+   * If the data is a "string" then you can specify which encoding is used. Default is "utf-8"
+   */
+  enc?: 'utf-8' | 'base16' | 'base64' | 'hex'
```
awoie, line 108: Since `enc` has a different semantics in the JWE world (i.e. encryption algorithm), I'd prefer to change the name of the property.
```suggestion
  encoding?: 'utf-8' | 'base16' | 'base64' | 'hex'
```

packages/core/src/types/IKeyManager.ts
```diff
@@ -81,6 +81,35 @@ export interface IKeyManagerDecryptJWEArgs {
...
+  /**
+   * The algorithm to use for signing.
+   * This must be one of the algorithms supported by the KMS for this key type.
+   */
+  alg?: string
+
+  /**
+   * Data to sign
+   */
+  data: string
```
awoie, line 103: Should this be rather a byte array? Then, the encoding parameter would be more powerful.

packages/kms-local/src/key-management-system.ts
```diff
@@ -68,21 +78,69 @@ export class KeyManagementSystem extends AbstractKeyManagementSystem {
...
+        }
+        if (typeof msgDomain !== 'object' || typeof msgTypes !== 'object') {
+          throw Error(`invalid_arguments: Cannot sign typed data. 'domain' and 'types' must be provided`)
+        }
+        const wallet = new Wallet(key.privateKeyHex)
+
+        const signature = await wallet._signTypedData(msgDomain, msgTypes, msg)
+        //HEX encoded string
+        return signature
+      }
```
awoie, line 135: Can we split this into separate methods for better readability and testability?

packages/core/src/types/IKeyManager.ts
```diff
@@ -81,6 +81,35 @@ export interface IKeyManagerDecryptJWEArgs {
...
 
+/**
+ * Input arguments for {@link IKeyManager.keyManagerSign | keyManagerSign}
+ * @public
+ */
+export interface IKeyManagerSignArgs {
+  /**
+   * Key ID
+   */
+  kid: string
```
awoie, line 92: i have probably looked at JOSE specs for too long, that is why i automatically think this is the value that goes into a JWE or JWS or JWK. if we used something for internal reference only but is different from what we'd put into the JOSE objects, then i would feel more comfortable if we just said `keyHandle` or `keyReference`.

packages/core/src/types/IIdentifier.ts
```diff
@@ -71,9 +71,14 @@ export interface IKey {
...
   /**
-   * Optional. Key metadata. Can be used to store auth data to access remote kms
+   * Optional. Key metadata. This should be used to determine which algorithms are supported.
    */
-  meta?: object | null
+  meta?: KeyMetadata | null
+}
+
+export interface KeyMetadata {
+  algorithms?: string[]
```
awoie, line 80: algorithms or crypto operations?

packages/core/src/types/IKeyManager.ts
```diff
@@ -81,6 +81,35 @@ export interface IKeyManagerDecryptJWEArgs {
...
+
+  /**
+   * The algorithm to use for signing.
+   * This must be one of the algorithms supported by the KMS for this key type.
+   */
+  alg?: string
```
awoie, line 98: This `alg` properly does not refer to the exhaustive IANA JOSE alg list, so we should use a different name, e.g., `keyOperation`. For example `eth_signTransaction` is something we only use internally and won't be registered in IANA. There will be also likely no JWT/JWS that uses alg= eth_signTransaction in the header.

awoie (changes requested):
more of my JOSE naming schizophrenia :)

packages/core/src/types/IKeyManager.ts
awoie: If we followed this approach, we could even go a little further and say, it is not called `Ed25519` or `EdDSA` but instead `sign_raw_eddsa` for example.

packages/core/src/types/IKeyManager.ts
mirceanis: If this were a byte array then the encoding param would not be needed. The encoding param is used to reason about the input data.
The constraint is that plugin methods need to be exposed over OpenAPI as well, and there is no support for byte arrays there.
I chose the string plus encoding approach to satisfy both open API and various input formats.

It's the responsibility of the `key-manager` plugin to convert the data string into a `UInt8Array` based on the encoding specified and to forward this to the KMS implementations, which do not have this OpenAPI constraint any more.

packages/core/src/types/IKeyManager.ts
mirceanis: good point.
My thinking here was to align to the `IKey` interface, but since that one has the same problem you're signaling, I can be more progressive and name this param better in advance

packages/core/src/types/IKeyManager.ts
mirceanis: I didn't want to add yet another name for the same thing, so to me it makes sense to call things by the same name as either IANA or ETH_RPC, or whatever the origin would be.

Using a different name instead of `alg` makes sense. I'll fix this

packages/core/src/types/IIdentifier.ts
mirceanis: I prefer algorithms, since they can represent both key operations (ex ECDH) or signature protocols (ex ES256K)

packages/core/src/types/IKeyManager.ts
mirceanis: done

packages/kms-local/src/key-management-system.ts
mirceanis: done

packages/core/src/types/IKeyManager.ts
mirceanis: done. renamed to `keyRef`

packages/core/src/types/IKeyManager.ts
mirceanis: renamed to `algorithm`

awoie (approved):
LGTM. It just strikes me a bit that 'Ed25519' and 'EdDSA' count as two separate algorithms although they are the same.

mirceanis:
> LGTM. It just strikes me a bit that 'Ed25519' and 'EdDSA' count as two separate algorithms although they are the same.

Yes, they are identical. I added both for compatibility.

PR #551, opened 2021-06-07
fix: speed up secp256k1 keygen

### LINKED ISSUE #549: secp256k1 key generation is too slow on react-native
**Bug severity**
3

**Describe the bug**
Since #529 got merged, secp256k1 key generation is very slow on react-native (tested on android).
The culprit seems to be the `pbkdf2` method that gets called by `@ethersproject/wallet#Wallet.createRandom()`

mirceanis:
fixes #549

Since #529 got merged, secp256k1 key generation is very slow on react-native (tested on android).
The culprit seems to be the `pbkdf2` method that gets called by `@ethersproject/wallet#Wallet.createRandom()`

This fix instantiates a `SigningKey` directly instead of deriving from a mnemonic, therefore sidestepping the call to pbkdf2

### DIFF
--- a/packages/kms-local/package.json
+++ b/packages/kms-local/package.json
@@ -12,6 +12,8 @@
     "@ethersproject/abstract-provider": "^5.1.0",
     "@ethersproject/abstract-signer": "^5.1.0",
     "@ethersproject/bytes": "^5.1.0",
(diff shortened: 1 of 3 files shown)


awoie (approved):
lgtm

PR #555, opened 2021-06-09
feat(key-manager): add method to compute a shared secret

### LINKED ISSUE #541: add `sharedSecret` method to `key-manager` and KMS abstracts
One of the core crypto operations that should be performed by the key management layer is computing a shared secret.
This would, for example, enable JWE [de]encrypters to use Veramo key management layer instead of requiring direct access to private key material.

* `key-manager` plugin needs to expose a `keyManagerSharedSecret()` method
* `AbstractKeyManagementSystem` class should have a new abstract method `computeSharedSecret()`

The basic API would be something like:
`computeSharedSecret(myKey: Partial<IKey>, theirKey: Partial<IKey>)`
using `Partial<>` since the user may wish to provide more key information that would ensure correctness.
Examples:
* `myKey.privateKeyHex` + `theirKey.publicKeyHex` + at least one key has `type` direct ECDH
* `myKey.type.isCompatible(theirKey.type) && (myKey.privateKeyHex || myKey.kid)` => KMS computes ECDH

The result is the RAW shared secret (scalar multiplication in the case of ECC).
To use this for symmetric encryption, one would need to hash it to arrive at a shared Symmetric Key

The `key-manager` plugin is expected to perform all (or most) of the checks regarding key compatibility and encoding.
The KMS implementations are expected to work mostly on raw Uint8Array data (and key handle in the case of `myKey`).

mirceanis:
## What's here

* `key-manager` plugin exposes a new method that computes a shared key (Diffie Hellman) between a secret key the agent manages and a public key of another party, returning a promise that resolves to the hex encoding string of the shared key.
```ts
async keyManagerSharedKey({secretKeyRef: string, publicKey: {type: string, publicKeyHex: string}}): Promise<string>`
```
The result is the RAW shared secret (scalar multiplication in the case of ECC).
To use this for symmetric encryption, one would need to hash it to arrive at a shared Symmetric Key
* To do this, there is a new requirement of the `AbstractKeyManagementSystem` class for a new abstract method `sharedKey()`
* The `kms-local` package implements this new `sharedKey` method
* `kms-local` supports `Ed25519` & `X25519` keys for computing the shared key. Keys are converted to `X25519` if they are `Ed25519`

closes #541

### DIFF
--- a/packages/core/src/types/IKeyManager.ts
+++ b/packages/core/src/types/IKeyManager.ts
@@ -202,11 +202,11 @@ export interface IKeyManager extends IPluginMethodMap {
   /**
    * Compute a shared secret with the public key of another party.
    * 
-   * This computes the raw shared key (the result of a Diffie-Hellman computation )
+   * This computes the raw shared secret (the result of a Diffie-Hellman computation)
    * To use this for symmetric encryption you MUST apply a KDF on the result.
    * 
    * @param args {@link IKeyManagerSharedKeyArgs} 
-   * @returns a `Promise` that resolves to a hex encoded shared key
+   * @returns a `Promise` that resolves to a hex encoded shared secret
    */
   keyManagerSharedSecret(args: IKeyManagerSharedSecretArgs): Promise<string>
 
--- a/packages/core/src/types/IKeyManager.ts
+++ b/packages/core/src/types/IKeyManager.ts
@@ -94,7 +94,7 @@ export interface IKeyManagerSignArgs {
   /**
    * The algorithm to use for signing.
    * This must be one of the algorithms supported by the KMS for this key type.
-   * 
+   *
    * The algorithm used here should match one of the names listed in `IKey.meta.algorithms`
    */
   algorithm?: string
@@ -110,6 +110,24 @@ export interface IKeyManagerSignArgs {
   encoding?: 'utf-8' | 'base16' | 'base64' | 'hex'
 
   [x: string]: any
+}
+
+/**
+ * Input arguments for {@link IKeyManager.keyManagerSharedSecret | keyManagerSharedSecret}
+ * @public
+ */
+export interface IKeyManagerSharedSecretArgs {
+  /**
+   * The secret key handle (`kid`)
+   * as returned by {@link IKeyManager.keyManagerCreate | keyManagerCreate}
+   */
+  secretKeyRef: string,
+
+  /**
+   * The public key of the other party.
+   * The `type` of key MUST be compatible with the type referenced by `secretKeyRef`
+   */
+  publicKey: Pick<IKey, 'publicKeyHex' | 'type'>
 }
 
 /**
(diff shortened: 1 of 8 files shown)


packages/core/src/types/IKeyManager.ts
```diff
@@ -181,6 +199,17 @@ export interface IKeyManager extends IPluginMethodMap {
    */
   keyManagerSign(args: IKeyManagerSignArgs): Promise<string>
 
+  /**
+   * Compute a shared secret with the public key of another party.
+   * 
+   * This computes the raw shared key (the result of a Diffie-Hellman computation )
```
awoie, line 205:
```suggestion
   * This computes the raw shared secret (the result of a Diffie-Hellman computation)
```

packages/core/src/types/IKeyManager.ts
```diff
@@ -181,6 +199,17 @@ export interface IKeyManager extends IPluginMethodMap {
...
+   * To use this for symmetric encryption you MUST apply a KDF on the result.
+   * 
+   * @param args {@link IKeyManagerSharedKeyArgs} 
+   * @returns a `Promise` that resolves to a hex encoded shared key
```
awoie, line 209:
```suggestion
   * @returns a `Promise` that resolves to a hex encoded shared secret
```

packages/key-manager/src/key-manager.ts
```diff
@@ -140,4 +142,20 @@ export class KeyManager implements IAgentPlugin {
...
+  async keyManagerSharedSecret(args: IKeyManagerSharedSecretArgs): Promise<string> {
+    const { secretKeyRef, publicKey } = args
+    const myKey = await this.store.get({ kid: secretKeyRef })
+    const theirKey = publicKey
+    if (
+      myKey.type === theirKey.type ||
+      (['Ed25519', 'X25519'].includes(myKey.type) && ['Ed25519', 'X25519'].includes(theirKey.type))
+    ) {
+    } else {
+      throw new Error('invalid_argument: the key types have to match to be able to compute a shared key')
```
awoie, line 156:
```suggestion
      throw new Error('invalid_argument: the key types have to match to be able to compute a shared secret')
```

packages/kms-local/src/key-management-system.ts
```diff
@@ -145,6 +145,24 @@ export class KeyManagementSystem extends AbstractKeyManagementSystem {
     throw Error(`not_supported: Cannot sign ${algorithm} using key of type ${key.type}`)
   }
 
+  async sharedSecret(args: { myKey: IKey; theirKey: Pick<IKey, 'type' | 'publicKeyHex'> }): Promise<string> {
+    const { myKey, theirKey } = args
+    let myKeyBytes = arrayify('0x' + myKey.privateKeyHex)
+    if (myKey.type === 'Ed25519') {
+      myKeyBytes = convertSecretKeyToX25519(myKeyBytes)
+    } else if (myKey.type !== 'X25519') {
+      throw new Error(`not_supported: can't compute shared key for type=${myKey.type}`)
```
awoie, line 154:
```suggestion
      throw new Error(`not_supported: can't compute shared secret for type=${myKey.type}`)
```

packages/kms-local/src/key-management-system.ts
```diff
@@ -145,6 +145,24 @@ export class KeyManagementSystem extends AbstractKeyManagementSystem {
...
+    }
+    let theirKeyBytes = arrayify('0x' + theirKey.publicKeyHex)
+    if (theirKey.type === 'Ed25519') {
+      theirKeyBytes = convertPublicKeyToX25519(theirKeyBytes)
+    } else if (theirKey.type !== 'X25519') {
+      throw new Error(`not_supported: can't compute shared key for type=${theirKey.type}`)
```
awoie, line 160:
```suggestion
      throw new Error(`not_supported: can't compute shared secret for type=${theirKey.type}`)
```

awoie (changes requested):
lgtm - just fixed some comments

PR #557, opened 2021-06-10
feat(key-manager): implement JWE functionality directly in `key-manager`

### LINKED ISSUE #556: [proposal] JWE functionality should be implemented on the `key-manager` layer
The `AbstractKeyManagementSystem` class is expected to perform some JWE encryption and decryption but it does not have all the prerequisites to deal with all JWE types (it may knowledge of key IDs as they appear in DID documents).

Since the `key-manager` already implements a `shared secret` method (see #541 and #555) it should be enough to do all the JWE packing on this layer, thus simplifying the implementations of `AbstractKeyManagementSystem`

mirceanis:
The KMS is only used to compute shared secret, it doesn't need to deal with JWE directly.
This PR moves the responsibility of createJWE and decryptJWE to the `key-manager` layer which eventually uses the `shared secret` functionality of the KMS to perform the same task.

fixes #556

### DIFF
--- a/packages/core/src/types/IKeyManager.ts
+++ b/packages/core/src/types/IKeyManager.ts
@@ -201,11 +201,11 @@ export interface IKeyManager extends IPluginMethodMap {
 
   /**
    * Compute a shared secret with the public key of another party.
-   * 
+   *
    * This computes the raw shared secret (the result of a Diffie-Hellman computation)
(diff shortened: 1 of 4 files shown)


awoie (approved):
lgtm - although i think the JWE functionality should be entirely removed at some point in favour of did-jwt

PR #561, opened 2021-06-12
Fixes #560, don't assume that message.data exists

### LINKED ISSUE #560: SDR message handler assumes that message.data exists
**Bug severity**
3

**Describe the bug**
The message handler for selective disclosure requests expects that message.data exists when evaluating the first if statement in the program. Message.data will not exist in certain error conditions, like an expired JWT or an unsupported message type.

Optional chaining on those variables needs to be added to cause that if statement to return false, instead of failing.

**Observed behaviour**
The program fails due to messagedata.type and message.data.claims being undefined

**Versions:**
 - Veramo: 1.2.2

rkreutzer:
Added optional chaining to the first if statement so that if message.data doesn't exist, the program doesn't fail.

### DIFF
--- a/packages/selective-disclosure/src/message-handler.ts
+++ b/packages/selective-disclosure/src/message-handler.ts
@@ -26,8 +26,8 @@ export class SdrMessageHandler extends AbstractMessageHandler {
     const meta = message.getLastMetaData()
 
     if (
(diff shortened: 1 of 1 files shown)


mirceanis (approved):
looks good

PR #575, opened 2021-06-17
feat: implement didcomm v2 packing/unpacking

### LINKED ISSUE #558: add didcomm pack method that accounts for key IDs and algorithms
required for #552

### LINKED ISSUE #559: Implement didcomm unpack method that resolved recipient key IDs
required for #553

mirceanis:
Implement `IDIDComm.packDIDCommMessage()` and `IDIDComm.unpackDIDCommMessage()` for plaintext, signed and encrypted (auth and anon) DIDComm.

fixes #558
fixes #559

This also adds a `getDIDComponentById` method to the `IResolver` plugin, which can be used to get a `VerificationMethod` or a `ServiceEndpoint` object form a DID document using an `id`.


packages/core/src/types/IResolver.ts
```diff
@@ -59,4 +65,61 @@ export interface IResolver extends IPluginMethodMap {
...
+   * @param args.section - Optional - the section of the DID Document to be used for dereferencing
+   *
+   * @returns a `Promise` containing the {@link did-resolver#VerificationMethod | VerificationMethod} or
+   *   {@link did-resolver#ServiceEndpoint | ServiceEndpoint}
+   *
+   * @throws `not_found:...` in case the fragment is not displayed in the DID document
+   *
+   * @beta
+   */
+  dereferenceDidUri(args: {
```
awoie, line 101: would it make sense to fully support `dereference` as described in DID Core: https://www.w3.org/TR/did-core/#did-url-dereferencing ? we should probably also add support for that in [did-resolver](https://github.com/decentralized-identity/did-resolver).

packages/did-comm/src/__tests__/did-comm.action-handler.test.ts
```diff
@@ -0,0 +1,150 @@
+import {
```
awoie, line 1: could we exclude this file from the PR since the changes are not final yet?

packages/did-comm/src/utils.ts
```diff
@@ -0,0 +1,306 @@
...
+
+export async function extractSenderEncryptionKey(
+  jwe: JWE,
+  context: IAgentContext<IResolver>,
+): Promise<Uint8Array | null> {
+  let senderKey: Uint8Array | null = null
+  const protectedHeader = decodeJoseBlob(jwe.protected)
+  if (typeof protectedHeader.skid === 'string') {
+    const senderDoc = await resolveDidOrThrow(protectedHeader.skid, context)
+    const sKey = (await context.agent.dereferenceDidUri({
```
awoie, line 112: Why do you need to resolve the DID Doc first? Couldn't dereferenceDidUri do the job itself?

awoie (commented):
i am not ready with the review yet but have already a few questions that need to be resolved before i would approve.

packages/did-comm/src/__tests__/did-comm.action-handler.test.ts
mirceanis: I will delete it since it wasn't actually testing did-comm related code

packages/did-comm/src/utils.ts
mirceanis: It's because otherwise resolve() would be called too many times and currently the cache in did-resolver is not well behaved.

packages/core/src/types/IResolver.ts
mirceanis: It would make sense, but the caching mechanism in did-resolver is not yet very well behaved so we might end-up calling resolve() too many times.

packages/did-resolver/src/resolver.ts
```diff
@@ -17,15 +25,59 @@ export class DIDResolverPlugin implements IAgentPlugin {
...
   }
+
+  /** {@inheritDoc @veramo/core#IResolver.dereferenceDidUri} */
+  async dereferenceDidUri({
+    didDocument,
+    didURI,
+    section,
+  }: {
+    didDocument: DIDDocument
+    didURI: string
```
awoie, line 53:
```suggestion
    didUrl: string
```

awoie: i also sometimes DID URI but i think the correct one is DID URL.

packages/did-resolver/src/resolver.ts
```diff
@@ -17,15 +25,59 @@ export class DIDResolverPlugin implements IAgentPlugin {
...
+  }: {
+    didUrl: string
+    options?: DIDResolutionOptions
+  }): Promise<DIDResolutionResult> {
     debug('Resolving %s', didUrl)
     return this.didResolver.resolve(didUrl, options)
   }
+
+  /** {@inheritDoc @veramo/core#IResolver.dereferenceDidUri} */
+  async dereferenceDidUri({
```
awoie, line 47:
```suggestion
  async dereferenceDidUrl({
```

packages/did-resolver/src/resolver.ts
mirceanis: got it.
It's fixed now

packages/did-resolver/src/resolver.ts
mirceanis: I renamed it to `getDIDComponentById`

packages/core/src/types/IResolver.ts
```diff
@@ -59,4 +65,61 @@ export interface IResolver extends IPluginMethodMap {
    * @public
    */
   resolveDid(args: ResolveDidArgs): Promise<DIDResolutionResult>
+
+  /**
+   * Dereferences a DID URI fragment and return the corresponding DID document entry.
```
awoie, line 70:
```suggestion
   * Dereferences a DID URI fragment and returns the corresponding DID document entry.
```

packages/did-comm/src/action-handler.ts
```diff
@@ -67,8 +130,355 @@ export class DIDComm implements IAgentPlugin {
...
+    let alg: string
+    if (signingKey.type === 'Ed25519') {
+      alg = 'EdDSA'
+    } else if (signingKey.type === 'Secp256k1') {
+      alg = 'ES256K'
+    } else {
+      throw new Error(
+        `not_supported: key of type ${signingKey.type} is not supported for JWS didcomm message`,
+      )
+    }
```
awoie, line 205: is this needed because the didcomm spec only allows a certain set of signature algs?

packages/did-comm/src/action-handler.ts
```diff
@@ -67,8 +130,355 @@ export class DIDComm implements IAgentPlugin {
...
+    // serialize flattened JWS JSON and return
+    return { message: JSON.stringify(packedMessage) }
+  }
+
+  private async packDIDCommMessageJWE(
+    args: IPackDIDCommMessageArgs,
+    context: IAgentContext<IDIDManager & IKeyManager & IResolver>,
+  ): Promise<IPackedDIDCommMessage> {
+    // 1. check if args.packing requires authentication and map sender key to skid
+    let senderECDH: ECDH | null = null
```
awoie, line 233: Would it make sense to have some validateDIDCommMessage that validates the basic message format, e.g., `from` can be a DID URL but no `#` allowed etc. This could be used for both pack functions obviously.

packages/did-comm/src/action-handler.ts
mirceanis: I think it's more a limitation of Veramo supported keys and algorithms.
For the moment there is a 1:1 correspondence between the key types and the JWA. If this changes, we would need to require an `alg` parameter to allow the user to specify

packages/did-comm/src/action-handler.ts
mirceanis: Sure, that makes sense. For the moment we rely on typescript to coerce the proper data types, but we don't do too much validation. We can add a validator.

packages/did-comm/src/types/IDIDComm.ts
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * Partially decodes a possible DIDComm message string to determine the {@link DIDCommMessageMediaType}
+   *
+   * @param IPackedDIDCommMessage - the message to be interpreted
+   * @returns the {@link DIDCommMessageMediaType} if it was successfully parsed
+   * @throws if the message cannot be parsed as DIDComm v2
+   *
+   * @beta
+   */
+  getDIDCommMessageMediaType(args: IUnpackDIDCommMessageArgs): Promise<DIDCommMessageMediaType>
```
awoie, line 33: `IUnpackDIDCommMessageArgs` seems to not match the param type.

mirceanis: `IUnpackDIDCommMessageArgs` is `{ message: string }`
I did not want to define a new virtually identical `*Args` type, especially since this method might be removed from the public plugin API.

packages/did-comm/src/action-handler.ts
awoie: i think for the beta version it should be fine to just demonstrate the raw functionality. since it is beta, it is subject to change anyways.

packages/did-comm/src/types/IDIDComm.ts
awoie: right, but would `IPackedDIDCommMessage` make more sense?

mirceanis: actually, yes. That's a good suggestion.

awoie (approved):
lgtm, it would be just nice to change the one argument of the getDIDCommMessageMediaType

PR #594, opened 2021-06-30
feat: add fake did method usable in tests

mirceanis:
I added this `did:fake` method to the integration test utilities.
This fake DID is resolved using the local agent database.

This PR also contains a fix for a bug in did-comm packing where some keys were getting converted twice from Ed25519 public to X25519 public, generating invalid packed messages.


simonas-notcat (approved):
Looks good

PR #606, opened 2021-07-07
Integration tests for add-didcomm-v2

simonas-notcat:


packages/did-comm/src/didcomm.ts
```diff
@@ -474,8 +474,8 @@ export class DIDComm implements IAgentPlugin {
 
     const services = didDoc.service?.filter(
       (service: any) =>
-        service.type == 'DIDCommMessaging' &&
-        service.accept?.includes('didcomm/v2'),
+        service.type == 'DIDCommMessaging' 
+        // && service.accept?.includes('didcomm/v2'),
```
mirceanis, line 478: @awoie is this `accept` field required for didcomm?

if so, I think we need to also do some refactoring of the way services are stored in our DB

mirceanis (approved):
Looks good to me.
I would like @awoie to comment on the `service.accept?.includes('didcomm/v2'),` question

packages/did-comm/src/didcomm.ts
awoie: yes, it is required

awoie: it is required because other people will use it to indicate didcommv2 messages and not aip2.0. we probably need to assume some default value but that is missing in the spec.

awoie: > If accept is not specified, the sender uses its preferred choice for sending a message to the endpoint.

awoie: so, if `accept` is missing, we assume didcommv2.

awoie (approved):
lgtm

packages/did-comm/src/didcomm.ts
mirceanis: this means we need to revise some aspects of did:ethr as well to support these service decorations

awoie: in the future probably. for now, we just assume it is `didcomm/v2`. although unlikely, but there could be other DID methods that do support other DIDComm profiles (e.g., AIP 2.0).

PR #610, opened 2021-07-08
feat: Added support for sending messages through simple HTTP-based transports

### LINKED ISSUE #469: Implement MessageHandler for handling DIDCommV2 envelopes
DIDCommV2 messages are identified by their cty and can have different forms:
- JWM (JWE) with ECDH-1PU --> needs to be decrypted with agents private key and needs to resolve public key of recipient
- JWM (JWS) with EdDSA/ES256K --> needs to be verified
- Plaintext

JWE and JWS are in JSON general serialization, NOT compact serialization

No threaded protocol support needed.

handleMessage should still emit the event validatedMessage in case everything went correctly

### LINKED ISSUE #552: Implement DIDComm Module for sending messages
- supports only JWE (auth/anoncrypt)
- includes: sending through transport manager
- packing should be done in #558

awoie:
Potentially closes #552 and potentially closes #469

Contains:
- Added transport interfaces to be used with DIDComm module. 
- Added simple implementation of an HTTP-based transport
- Implemented sending DIDComm messages

Add a list of supported transports to the DIDComm plugin:

```js
new DIDComm([new DIDCommHttpTransport()])
```

Then, use the DIDComm module to send messages to a DID:

```js
const message = {
  type: 'test',
  to: receiver.did,
  from: sender.did,
  id: 'test',
  body: { hello: 'world' },
}
const packedMessage = await agent.packDIDCommMessage({
  packing: 'authcrypt',
  message,
})
const result = await agent.sendDIDCommMessage({
  messageId: '123',
  packedMessage,
  recipientDidUrl: receiver.did
})
```

### DIFF
--- a/__tests__/localAgent.test.ts
+++ b/__tests__/localAgent.test.ts
@@ -159,6 +159,7 @@ const setup = async (options?: IAgentOptions): Promise<boolean> => {
       new DIDDiscovery({
         providers: [new AliasDiscoveryProvider(), new ProfileDiscoveryProvider()],
       }),
+      ...options?.plugins || [],
     ],
   })
   return true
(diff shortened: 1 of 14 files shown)


mirceanis (approved):
looks great

PR #625, opened 2021-07-20
fix(cli): export recent methods from CLI local and remote

mirceanis:
HOTFIX: The recently added plugin methods were not usable with CLI.
This PR fixes that.


simonas-notcat (approved):

jasonhealy (commented):
Looks good

PR #627, opened 2021-07-21
feat: ceramic did provider

simonas-notcat:
# DID Provider for ceramic

## Example usage

```typescript
import { VeramoDidProvider } from '@veramo/ceramic'
import { DID } from 'dids'
import KeyDidResolver from 'key-did-resolver'
import CeramicClient from '@ceramicnetwork/http-client'
import { TileDocument } from '@ceramicnetwork/stream-tile'
import { agent } from './setup.ts'

const API_URL = "https://ceramic-clay.3boxlabs.com"

const alice = await agent.didManagerGetOrCreate({
  alias: 'alice',
  provider: 'did:key'
})
const provider = new VeramoDidProvider(agent, alice.did)

const resolver = { ...KeyDidResolver.getResolver() }
const did = new DID({ provider, resolver })
await did.authenticate()

const ceramic = new CeramicClient(API_URL)
ceramic.setDID(did)

const content = { hello: 'from veramo' }
const doc = await TileDocument.create(ceramic, content)
```


packages/ceramic/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2019 Consensys AG
```
mirceanis, line 189:
```suggestion
   Copyright 2021 Consensys AG
```

package.json
```diff
@@ -72,6 +72,12 @@
     ]
   },
   "dependencies": {
+    "@ceramicnetwork/core": "^1.2.0",
+    "@ceramicnetwork/http-client": "^1.0.7",
+    "@ceramicnetwork/stream-tile": "^1.0.7",
+    "dids": "^2.4.0",
+    "key-did-provider-ed25519": "^1.1.0",
+    "key-did-resolver": "^1.2.1",
```
mirceanis, line 80: looks like some dependencies are not placed properly.
I believe these should be in the `@veramo/ceramic` package.

mirceanis (changes requested):
looks good after a first pass.

Some dependencies seem to be installed in the wrong place. Please take a look.
We should be able to implement `decrypt...` as well. Would that be enough to unskip that skipped test?

package.json
simonas-notcat: These dependencies are needed for integration tests. They don't belong in `@veramo/ceramic` package

PR #661, opened 2021-08-11
feat(key-manager): move private key storage to kms-local

### LINKED ISSUE #539: add key storage capabilities directly to kms-local
It should be up to the KMS(`AbstractKeyManagementSystem`) implementations to decide how (and if) they store private key material.
KMS implementations should not rely on a `privateKeyHex` field being supplied by the `key-manager` plugin.

Constraints:
* the `key-manager` plugin (as any other agent plugin) should never be able to reach private key material directly.
* the `kms-local` package would leverage the same storage as before, even after the migration of key storage responsibility.
* the core packages should allow static configurations (like `agent.yml` for CLI)
* migration to the new KMS should be seamless

### LINKED ISSUE #540: remove private key information access from the top-level `key-manager` plugin
It should be up to the KMS(`AbstractKeyManagementSystem`) implementations to decide how (and if) they store private key material.
KMS implementations should not rely on a `privateKeyHex` field being supplied by the `key-manager` plugin.
After #539 is implemented, the `key-manager` should not require a `KeyStore` any more

Constraints:
* the `key-manager` plugin (as any other agent plugin) should never be able to reach private key material directly.
* the core packages should still allow static configurations (like `agent.yml` for CLI)
* migration to the new KMS should be seamless

### LINKED ISSUE #680: [proposal] Better key and DID import
I feel on shaky grounds when importing keys (with DID) into Veramo. I dont know if i do it correctly and getting the keys with all information needed is a hassle.

A function that takes PK and maybe keytype (if it cant be derived).

Currently we are saving all key and DID info in .evn variables then doing an import of those if we dont find those keys into the database. We are also trying to make sure we are only running on set of keys on the agent (running on a server).

**Additional context**

26-08-2021 at 08 26 09

mirceanis:
### What's new

* `@veramo/key-manager` no longer holds private key material
* `@veramo/kms-local` gets its own `PrivateKeyStore`
* Keys and DIDs are easier to import.

```typescript
const managedKeyInfo = await agent.keyManagerImport({
        kms: 'local',
        type: <TKeyType>'Secp256k1',
        privateKeyHex: '31d1ec1...5f8655ec63',
        kid: 'this is optional'
        // all other properties are derived by KMS
      })
```
---
### How to upgrade

If you're already working with Veramo and wish to upgrade **existing agents** to veramo 3.0, you'll have to make some changes to your configuration, depending on how you're using the framework.

It boils down to these 3 steps:
1. Update your database connection to use `migrations`
2. Remove the `SecretBox` parameter from `KeyManager`
3. Add a `PrivateKeyStore` parameter to `KeyManagementSystem` with a `SecretBox` that you were using before with `KeyManager` (and keep the same encryption key) 

### Typescript app config changes

If your agents are part of a typescript app, these changes will look like this:

```typescript
import {
  // ...
  migrations, // 0. import default veramo migrations
  PrivateKeyStore
} from '@veramo/data-store'

// 1. update your connection config to use migrations
dbConnection = createConnection({
    // ...
    synchronize: false, // switch this to false
    migrations,         // add default veramo migrations
    migrationsRun: true,// add this flag
  })

const agent = createAgent<>({
// ...
    plugins: [
      // 2. key manager: remove SecretBox param
      new KeyManager({
        store: new KeyStore(dbConnection),
        kms: {
          // 3. kms-local: add PrivateKeyStore with SecretBox
          local: new KeyManagementSystem(new PrivateKeyStore(dbConnection, new SecretBox(secretKey))),
        },
      }),
      // ...
    ]})
```

### CLI config changes

```yml
# 0. update the version for your config file
version: 3.0
# ...

# 1. update your database connection to use migrations
dbConnection:
  $require: typeorm?t=function#createConnection
  $args:
    - type: sqlite
      database:
        $ref: /constants/databaseFile
      synchronize: false  # switch off synchronize
      migrationsRun: true # turn on migrations
      migrations:         # use veramo default migrations
        $require: '@veramo/data-store?t=object#migrations'
      logging: false
      entities:
        $require: '@veramo/data-store?t=object#Entities'

# then update your keyManager config:
keyManager:
  $require: '@veramo/key-manager#KeyManager'
  $args:
    - store:
        $require: '@veramo/data-store#KeyStore'
        $args:
          - $ref: /dbConnection
          # 2. remove the SecretBox argument from KeyManager
      kms:
        local:
          $require: '@veramo/kms-local#KeyManagementSystem'
          $args:
            - $require: '@veramo/data-store#PrivateKeyStore'
              $args:
                - $ref: /dbConnection
                # 3. add the SecretBox argument from KeyManager to your PrivateKeyStore
                - $require: '@veramo/kms-local#SecretBox'
                  $args:
                    - $ref: /constants/secretKey
```

---

fixes #539
fixes #540
fixes #680 

BREAKING CHANGE: `keyManagetGet` no longer returns private key data
BREAKING CHANGE: `KeyStore` no longer requires a `SecretBox`
BREAKING CHANGE: `KeyManagementSystem` needs a `PrivateKeyStore`
BREAKING CHANGE: @veramo/cli configuration version update to 3.0


__tests__/migration.test.ts
```diff
@@ -0,0 +1,142 @@
...
+import { DIDManager } from '../packages/did-manager'
+import { FakeDidProvider, FakeDidResolver } from './utils/fake-did'
+import fs from 'fs'
+
+jest.setTimeout(30000)
+
+const databaseBeforeFile = __dirname + '/fixtures/local-database-before-3.0.sqlite'
+const databaseFile = __dirname + '/migrated.database.sqlite'
+const infuraProjectId = '5ffc47f65c4042ce847ef66a3fa70d4c'
+const secretKey = '29739248cad1bd1a0fc4d9b75cd4d2990de535baf5caadfdf8d8f86664aa830c'
```
kdenhartog, line 33: Is this meant to be a secretKey with infura? If it is and it's actually meant to be secret I'd suggest passing it in as an environment var through a config file so it's not being revealed in commit history

mirceanis: This key is meant to decrypt some data in the static DB used during tests.
That is a very good observation. I will try to incorporate it for best practice.

simonas-notcat (approved):
Great job @mirceanis !

PR #679, opened 2021-08-25
feat(data-store): initialize DB using migrations

### LINKED ISSUE #676: change to true DB migration (away from synchronize that breaks in production)
@trentlarson commented on [Mon Apr 26 2021](https://github.com/uport-project/veramo-website/pull/68)

See 'synchronize' explanation here: https://github.com/typeorm/typeorm/blob/master/docs/connection-options.md

Everything else appears to work in real apps (as long as https://github.com/uport-project/veramo/pull/480 is included for mobile), so adding this will save people heartburn as they create their apps.

---

@mirceanis commented on [Thu Apr 29 2021](https://github.com/uport-project/veramo-website/pull/68#issuecomment-829005677)

This is very interesting. My gut tells me that this migration logic should be part of the `@veramo/data-store` plugin directly. What are your thoughts about this?

---

@mirceanis commented on [Thu Jul 01 2021](https://github.com/uport-project/veramo-website/pull/68#issuecomment-872066951)

@simonas-notcat perhaps you can share some wisdom here :)

---

@mirceanis commented on [Fri Aug 20 2021](https://github.com/uport-project/veramo-website/pull/68#issuecomment-902652774)

taking initial steps to promote migrations instead of `synchronized`: https://github.com/uport-project/veramo/pull/661

mirceanis:
fixes #676

## BREAKING CHANGES

The `@veramo/data-store` package relies on `typeorm` as a database abstraction.
Typeorm has a connection flag `synchonize` which bootstraps the database along with schema and relations based on a set of `Entities` (annotated typescript classes).
This is very handy for fast development iterations but it is **not recommended for production** use because there is too much ambiguity possible when the `Entities` change, and there is a risk of data loss.
The recommended way to do things is to use the `migrations` mechanism. It allows you to migrate to new database schemas when necessary, and even customize the database to your own needs.

**Going forward, this is the mechanism we will be recommending for connections.**

To adopt this, there are some small changes needed for your agent configuration:

### new `@veramo/cli` configuration:

If you're already running an agent using the `@veramo/cli` package, please update the `dbConnection` section of your configuration like so:

```yml
# Data base
dbConnection:
  $require: typeorm?t=function#createConnection
  $args:
    - type: sqlite
      database:
        $ref: /constants/databaseFile
      synchronize: false    # set this to false
      migrationsRun: true   # set this to true
      migrations:           # and import the default migrations from veramo
        $require: '@veramo/data-store?t=object#migrations' 
      logging: false
      entities:
        $require: '@veramo/data-store?t=object#Entities'
```

If you haven't edited your configuration before, you can simply create a new config file with the `veramo config create` command.

### new JS/TS configuration:

```typescript
import { Entities, migrations, /*...*/ } from '../packages/data-store/src'
import { createConnection } from 'typeorm'
// ...
const dbConnection = createConnection({
   // ...
    synchronize: false,     // set this to false
    migrations: migrations, // import the default migrations from veramo
    migrationsRun: true,    // enable this flag
    logging: false,
    entities: Entities,
  })

// the agent configuration doesn't actually change, only the `dbConnection` from above.
agent = createAgent<>({
  plugins: [
    // ... use the dbConnection from above with the plugins that require it
    new DataStore(dbConnection),
    new DataStoreORM(dbConnection),
  ]
})
```

---
#### PR Tasks
- [X] create initial migrations using `querryRunner` API
  - align cascade options
  - align `nullable: true` with respective Entity properties
- [X] add test for DB initialized using migrations
  - decouple test data. Test suites should be able to run in isolation.
- [x] add smoke tests using migrated data ( see [initial.migration.test.ts](https://github.com/uport-project/veramo/pull/679/files#diff-d546b162564cc8e878144880fe457f0e1092455a1d0503f92950d9b6dde86290))
  * [x] get key
  * [x] get identifier with keys and services
  * [x] get credential
  * [x] get credential by claim
  * [x] get presentation
  * [x] get message with credentials and presentation
- [x] run tests using different DB drivers (sqlite & postgres)
  * [x] automate tests using postgres (see [dbInitTests](https://github.com/uport-project/veramo/pull/679/files#diff-cccff934f848572322c0f6c43203ab6504b532f69ce3f5529002c8e0d2c86b4b))
- [x]  check for data loss if migrations are NOT used
  * app will fail to start because of broken FOREIGN KEY constraints if you try to start without `migrations` and with `synchronize: true`


simonas-notcat (approved):
Looks great. Awesome work @mirceanis !

PR #713, opened 2021-09-29
feat: MessagingRouter `save` option

### LINKED ISSUE #712: [proposal] Add a flag to MessagingRouterOptions for automatically saving messages
There are situations where I'm interested only in handling `validatedMessage` event. But I don't want the message to be automatically saved.

There should be an option in `MessagingRouterOptions` - `save: boolean`

https://github.com/uport-project/veramo/blob/03343b4d902a8fb904d61869f50fe7ee72521112/packages/remote-server/src/messaging-router.ts#L34-L38

simonas-notcat:
Fixes #712

### DIFF
--- a/packages/remote-server/src/messaging-router.ts
+++ b/packages/remote-server/src/messaging-router.ts
@@ -16,6 +16,11 @@ export interface MessagingRouterOptions {
     type: string
     value?: string
   }
+
+  /**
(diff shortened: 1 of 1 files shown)


mirceanis (approved):
Looks good.
We should be removing the automatic saving from most places anyway :)

PR #725, opened 2021-10-21
fix(data-store): add support for entityPrefix

### LINKED ISSUE #724: Data-store migrations don't support entityPrefix
**Bug severity**
3

**Describe the bug**
My TypeORM connection is using the entityPrefix parameter to prepend characters to the database table names, so as not to confuse with other tables in the system. After moving to v3 and using the migrations approach in the database connection, new, empty tables are created without the entity prefix, as well as the private_key table.

My workaround was to hard-code the prefix on all table names in the migrations code.

**To Reproduce**
1. On an existing app that was running v2 using an entityPrefix on the database connection, bump to v3 and use migrations

**Observed behaviour**
A new set of database tables are created, alongside the tables that have the entity prefix. Also, the private_key table is created without the entity prefix.
   
**Expected behaviour**
The entityPrefix should be prepended to all tables in the migrations files.

 - Veramo: 3.0
 - Ionic/angular/capacitor app
 - Node Version 14.16.1

mirceanis:
fixes #724

I use the `Entity` metadata that TypeORM provides (which includes the `entityPrefix` configured on the connection) to decide on the database table names for migrations.

> Users that use `entityPrefix` MUST skip the upgrade from 2.x to 3.0.0 and jump straight to the version that includes this fix!

I merged the pre-migration database fixtures into a single db and added another fixture to test migrations when `entityPrefix` is used.

This PR also contains some doc fixes for some core types and a workaround for a bug in the api-extractor toolchain that introduces some whitespace in `{@link URLs}`.
Also, the default `infuraProjectId` used in testing is updated.

### DIFF
--- a/packages/data-store/src/migrations/createPrivateKeyStorage.ts
+++ b/packages/data-store/src/migrations/createPrivateKeyStorage.ts
@@ -63,6 +63,12 @@ export class CreatePrivateKeyStorage1629293428674 implements MigrationInterface 
   }
 
   async down(queryRunner: QueryRunner): Promise<void> {
+    function getTableName(givenName: string): string {
+      return (
+        queryRunner.connection.entityMetadatas.find((meta) => meta.givenTableName === givenName)?.tableName ||
+        givenName
+      )
+    }
     // 1. add old column back
     debug(`adding back privateKeyHex column to key table`)
     await queryRunner.addColumn(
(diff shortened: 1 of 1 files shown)


rkreutzer (commented):
Line 90 in createPrivateKeyStorage needs the prefix

PR #777, opened 2021-12-09
fix(remote-server): web-did-doc-router

simonas-notcat:
Properly checking if default services are provided


mirceanis (approved):
Looks great!

PR #800, opened 2022-01-21
chore: upgrade ethr-did-resolver

simonas-notcat:


mirceanis (approved):
looks good, thanks for propagating this update

PR #807, opened 2022-02-02
fix: add missing uuid dependency

simonas-notcat:
https://github.com/uport-project/veramo/blob/28c35e187caa9e5d56e149e5be220f6c9e14e0fb/packages/key-manager/src/memory-key-store.ts#L8


mirceanis (approved):
looks good

PR #814, opened 2022-02-15
feat: credentialStatus support

tsabolov:
In order to support credential status checks a `checkStatus` function must be passed to vc.js's method `verifyCredential`. A signature of `verifyCredentialLD` method was changed to support the alternative call with `checkStatus` member.

### DIFF
--- a/packages/credential-ld/plugin.schema.json
+++ b/packages/credential-ld/plugin.schema.json
@@ -331,6 +331,10 @@
               "type": "boolean",
               "description": "Set this to true if you want the '@context' URLs to be fetched in case they are not pre-loaded.",
               "default": false
+            },
+            "checkStatus": {
+              "description": "Optional function for checking credential status if `credentialStatus` is present on the credential",
+              "default": "undefined"
             }
           },
           "required": [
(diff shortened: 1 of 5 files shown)


mirceanis:

We had a slightly different idea for how `credentialStatus` fields could be handled and checked.
We were thinking that a Veramo plugin that exposes a method such as `checkStatus(credential, options)` would be a better option.
This method can be called automatically during verification, or directly by the user.

The first reason for this is that there can be multiple status methods (`RevocationList2021`, `EthrStatusRegistry2019`, `CredentialList2017`, etc..) and they have different requirements and implementations. From this perspective, this functionality is very similar to the DIDResolver plugin, which is also automatically used during credential verification.
These implementations should be "installed" during initialization and then the user doesn't have to interpret the `credentialStatus` field manually to decide which implementation fits. This could be done by the plugin that orchestrates these checks.

Also, all the plugin methods that we have so far in this core repository have serializable arguments. This enables us to call these methods over OpenAPI. We don't expect this requirement to hold for all possible plugins but we wish to keep it true for the core plugins.

Does this make sense? What do you think?

tsabolov:
Thank you for the detailed explanation. That makes a lot of sense, and I think that approach is more in line with what's already implemented.

Have you already started working on the feature? If yes, I would join the discussion/implementation :)

mirceanis:
See https://github.com/uport-project/credential-status/pull/39 for some progress on the credential format support

mirceanis:
@tsabolov do you know of any implementations of `credentialStatus` methods that you can point to?

tsabolov:
Sorry @mirceanis, somehow missed your message. Need to review my GitHub notifications settings.
We are going to use our own [`RevocationStatus2021LDAP`](https://spherity.github.io/vc-status-2021-ldap/) for one of our projects. It's still a draft, so I'd not include it into the list yet.

PR #819, opened 2022-02-23
Feature/data store json

mirceanis:
This PR adds a plugin that implements the `IDataStore` and `IDataStoreORM` interfaces using a JSON tree as the backing mechanism.
There is an optional callback that implementors can use to track updates to the data being stored, and persist it to some media, otherwise it is all stored in memory.

* [X] implement JSON tree based `IDataStore` (data saving)
* [X] implement JSON tree based `IDataStoreORM` (data querying)
* [X] add integration test using `DataStoreJson` saved to a file for all data needs.
* [X] adapt react-sample browser tests to use `DataStoreJson` to increase the test coverage
* [x] move `IDataStore` and `IDataStoreORM` interface definitions to `@veramo/core` package
* [x] Add notes about requirement to use the same data storage to make queries work.
* [ ] Adapt react-sample to use session storage to persist agent data using data-store-json
* [ ] rename `IDataStoreORM` to `IDataStoreQuery`, marking it BREAKING CHANGE


packages/data-store-json/src/identifier/did-store.ts
```diff
@@ -0,0 +1,93 @@
...
+    provider?: string
+  }): Promise<IIdentifier> {
+    let where: { did?: string; alias?: string; provider?: string } = {}
+
+    if (did !== undefined && alias === undefined) {
+      where = { did }
+    } else if (did === undefined && alias !== undefined && provider !== undefined) {
+      where = { alias, provider }
+    } else {
+      throw Error('[veramo:data-store:identifier-store] Get requires did or (alias and provider)')
```
simonas-notcat, line 42:
```suggestion
      throw Error('[veramo:data-store-json:identifier-store] Get requires did or (alias and provider)')
```

simonas-notcat (approved):

packages/data-store-json/src/identifier/did-store.ts
mirceanis: good catch :)

PR #844, opened 2022-03-30
chore: Regenerate did-comm plugin schema

nickreynolds:
* Generating schema on Windows machine added extra escaped character. Regenerated on unix to remove.


mirceanis (approved):
It's worth having a windows view of the project as well.
Thanks for spotting and fixing this

PR #868, opened 2022-05-06
fix(jsonld-context): Removing fs dependency for JSON LD default contexts

### LINKED ISSUE #837: [BUG] credential-ld module uses require('fs') which limits the platforms where it can be used
This is used to load `@context` objects asynchronously.
This should be replaced by another mechanism that also works in browser environments.

italobb:
Closes #837

Using async import instead of require('fs').


simonas-notcat (changes requested):
I think you also need to make sure that `contexts/*.json` files are included in the package https://github.com/uport-project/veramo/blob/881e96ae148b25622ec1db58d36650cfcd6ff934/packages/credential-ld/package.json#L38

mirceanis (approved):
This looks good, did you get a chance to check it in a browser environment too?

italobb:
@mirceanis That's my next step. I'm trying to reuse the `__browser_tests__` to implement that test. It was broken and I already fixed it, but I'm trying to understand how it works.

PR #874, opened 2022-05-10
Credential status plugin for Veramo

tsabolov:
Hi folks. This is our another (first one was [here](https://github.com/uport-project/veramo/pull/814)) try of implementing support for the [`credentialStatus` field](https://www.w3.org/TR/vc-data-model/#status) of W3C credentials.
In fact, it's quite simple - just wrapping the [`credential-status`](https://github.com/uport-project/credential-status) library.

Any feedback is welcome and appreciated.

### DIFF
--- a/packages/credential-ld/src/ld-credential-module.ts
+++ b/packages/credential-ld/src/ld-credential-module.ts
@@ -145,18 +145,23 @@ export class LdCredentialModule {
     fetchRemoteContexts: boolean = false,
     context: IAgentContext<IResolver>,
   ): Promise<boolean> {
+    const checkStatusFn =
+      typeof context.agent.checkCredentialStatus === 'function'
+        ? async ({ credential }: any) => {
+            const result = await context.agent.checkCredentialStatus({
+              credential,
+            })
+            return { verified: !result.revoked }
+          }
+        : void 0
+
     const result = await vc.verifyCredential({
       credential,
       suite: this.ldSuiteLoader.getAllSignatureSuites().map((x) => x.getSuiteForVerification()),
       documentLoader: this.getDocumentLoader(context, fetchRemoteContexts),
       purpose: new AssertionProofPurpose(),
       compactProof: false,
-      checkStatus: async ({ credential }: any) => {
-        const result = await context.agent.checkCredentialStatus({
-          credential,
-        });
-        return { verified: !result.revoked };
-      },
+      checkStatus: checkStatusFn,
     })
 
     if (result.verified) return true
--- a/packages/credential-ld/src/ld-credential-module.ts
+++ b/packages/credential-ld/src/ld-credential-module.ts
@@ -151,6 +151,12 @@ export class LdCredentialModule {
       documentLoader: this.getDocumentLoader(context, fetchRemoteContexts),
       purpose: new AssertionProofPurpose(),
       compactProof: false,
+      checkStatus: async ({ credential }: any) => {
+        const result = await context.agent.checkCredentialStatus({
+          credential,
+        });
+        return { verified: !result.revoked };
+      },
     })
 
     if (result.verified) return true
--- a/packages/credential-status/src/__tests__/credential-status.test.ts
+++ b/packages/credential-status/src/__tests__/credential-status.test.ts
@@ -1,0 +1,73 @@
+import { createAgent } from '../../../core/src'
+import { CredentialStatusPlugin } from '../credential-status'
+import { DIDDocument } from 'did-resolver'
+
+describe('@veramo/credential-status', () => {
+  const referenceCredential = {
+    credentialStatus: {
+      type: 'ExoticStatusMethod2022',
+      id: 'some-exotic-id',
+    },
+  }
+  const referenceDoc = {} as DIDDocument
+
+  it('should check the credential status', async () => {
+    const expectedResult = {}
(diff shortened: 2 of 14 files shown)


packages/credential-status/src/credential-status.ts
```diff
@@ -0,0 +1,22 @@
+import { IAgentContext, IAgentPlugin, IPluginMethodMap } from '@veramo/core';
+import { Status, StatusMethod } from 'credential-status';
+import { ICheckCredentialStatusArgs } from './types';
+
+export class CredentialStatusPlugin implements IAgentPlugin {
+  private readonly status: Status;
+  readonly methods: IPluginMethodMap;
```
mirceanis, line 7: I think this should be ICheckCredentialStatus

```suggestion
  readonly methods: ICheckCredentialStatus;
```

packages/credential-ld/src/ld-credential-module.ts
```diff
@@ -151,6 +151,12 @@ export class LdCredentialModule {
       documentLoader: this.getDocumentLoader(context, fetchRemoteContexts),
       purpose: new AssertionProofPurpose(),
       compactProof: false,
+      checkStatus: async ({ credential }: any) => {
+        const result = await context.agent.checkCredentialStatus({
```
mirceanis, line 155: This will fail if the status plugin is not installed.
We'll have to find a way to check first.

packages/credential-status/src/__tests__/credential-status.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  const referenceCredential = {
+    credentialStatus: {
+      type: 'ExoticStatusMethod2022',
+      id: 'some-exotic-id',
+    },
+  }
+  const referenceDoc = {} as DIDDocument
+
+  it('should check the credential status', async () => {
+    const expectedResult = {}
```
mirceanis, line 15: It's probably a good idea to predefined the number of assertions since we're dealing with async methods.
```suggestion
    expect.assertions(3)
    const expectedResult = {}
```

mirceanis (changes requested):
This is wonderful, thank you for the contribution!

I noted some minor changes needed.

Now the pressure is higher to get `credential-status` up to date and to also hook things up to real-world status methods too 😅

packages/credential-status/src/credential-status.ts
tsabolov: Makes sense. Fixed.

packages/credential-ld/src/ld-credential-module.ts
tsabolov: Fair enough :D Added a check. Please take a look at the solution.

mirceanis:
Everything looks good on this end but I think there might be something wrong with the credential-status library that makes the build fail.

tsabolov:
The problem is that `credential-status` module is published into NPM as typescript sources, not the compiled version. I added a workaround for that into the [tsconfig](https://github.com/uport-project/veramo/blob/cbee8437dc9c23a5627b9d344cdcf7248aaf6391/packages/credential-status/tsconfig.json#L8-L11) of the plugin. During the integration tests the common tsconfig is being used, which does not contain the workaround for the library.

image

I guess, the only way to fix the issue is to publish `credential-status` library on NPM as a compiled JS.

mirceanis:
> The problem is that `credential-status` module is published into NPM as typescript sources, not the compiled version. I added a workaround for that into the [tsconfig](https://github.com/uport-project/veramo/blob/cbee8437dc9c23a5627b9d344cdcf7248aaf6391/packages/credential-status/tsconfig.json#L8-L11) of the plugin. During the integration tests the common tsconfig is being used, which does not contain the workaround for the library.
> 
> image
> 
> I guess, the only way to fix the issue is to publish `credential-status` library on NPM as a compiled JS.

I just published a fix for credential-status to include the compiled code (v2.0.1).

It should be OK now and you probably also don't need the workarounds anymore either :)

tsabolov:
> I just published a fix for credential-status to include the compiled code (v2.0.1).
> 
> It should be OK now and you probably also don't need the workarounds anymore either :)

The library is published as an ESM module (with `import`s). My TypeScript application is compiled into CommonJS modules, so it's unable to use the library. I see that it's built with `microbundle` which I guess by default compiles/bundles for browsers and ESM. Could you change to CommonJS output and re-publish the lib once again?
Sorry :)

mirceanis:
> Could you change to CommonJS output and re-publish the lib once again? Sorry :)

I'm away from my laptop these days because I'm on paternity leave. I've been doing the best I can on my phone but I won't be able to test such a change properly.

 If you know exactly what must be done please raise a PR on credential-status too and I'll be able to review and merge.

Microbundle should know how to interpret package.json and produce both kinds of outputs and suspect the only change needed is in the library's package.json file, but feel free to propose any change necessary to get things moving there.

tsabolov:
Ah, I prepared a [PR](https://github.com/uport-project/credential-status/pull/48) to `credential-status` but you were faster than me :D 
I guess that this PR's build should be somehow restarted.

mirceanis:
> Ah, I prepared a [PR](https://github.com/uport-project/credential-status/pull/48) to `credential-status` but you were faster than me :D I guess that this PR's build should be somehow restarted.

I guess you can bump the credential-status dependency here to 2.0.2 to import the right one. This should also trigger the build again

Issue #857, opened 2022-04-20
data-store-json fails to clone objects in a browser environment


simonas-notcat:
When trying to use `@veramo/data-store-json` in a react app, this error is thrown:

```
Failed to execute 'structuredClone' on 'Window': async (oldState, newState) => {
      this.save(newState);
    } could not be cloned.
```

It looks like the browser is executing a built-in `structuredClone` method instead of using the polyfilled one.

PR #885, opened 2022-05-16
fix(data-store-json): structuredClone

simonas-notcat:
Relates to https://github.com/uport-project/veramo/issues/857


mirceanis:
Is it possible that #857 is being caused by having the storage object somehow contain `window` and thus causing the failure?

simonas-notcat:
@mirceanis This is already what is happening under the hood:
https://github.com/ungap/structured-clone/blob/main/cjs/index.js#L20

I'm simply forcing this behavior.

mirceanis:
> @mirceanis This is already what is happening under the hood

Haha, you are right, I did not think to look.

mirceanis (approved):
Looks good

PR #888, opened 2022-05-20
feat: Create DIDComm JWE with multiple recipients

inevolin:
**Problem:** We want to create a JWE for multiple recipients (ie.: different DID docs), by means of including these as recipients as well. Currently the JWE's recipients are only coming from the beneficiary's DID doc (keyAgreements).

**Proposal:** Extending the DIDCommMessage with a `bcc` field (~blind carbon copy) which is an array of additional beneficiary DIDs. Our proposed code resolves these into DIDdocs, and adds their keyAgreements into the recipients as well.

We don't care about sending DIDComm messages to the `bcc` recipients right now, but anyone can extend this logic in the future if so desired.

---

**Testing:** Proper unit test(s) can be added if necessary. But here's a POC test script for our use case:

```ts
import { IIdentifier } from "../../veramo/packages/core/build";
import { agent } from "./veramo/setup";

async function main() {
  const identifiers = await agent.didManagerFind();
  jwe(...identifiers);
}
async function jwe(
  originator?: IIdentifier,
  beneficiary1?: IIdentifier,
  beneficiary2?: IIdentifier
) {
  originator ||= await agent.didManagerCreate({
    kms: "local",
    provider: "did:key",
  });
  beneficiary1 ||= await agent.didManagerCreate({
    kms: "local",
    provider: "did:key",
  });
  beneficiary2 ||= await agent.didManagerCreate({
    kms: "local",
    provider: "did:key",
  });
  // these beneficiaries currently "live" in the same local database, but they can live anywhere apart.

  const message = {
    type: "test",
    from: originator.did,
    to: originator.did,
    bcc: [beneficiary1.did, beneficiary2.did],
    id: "test",
    body: { hello: "world" },
  };

  // create JWE
  const emsg = await agent.packDIDCommMessage({
    packing: "authcrypt",
    message,
  });
  console.dir(emsg, { depth: 10 });

  // decrypt JWE
  const msg = await agent.unpackDIDCommMessage({ message: emsg.message });
  console.dir(msg, { depth: 10 });
  // decryption is possible as long as we know the full `did:key` one of any of the recipients.
}

main().catch((e) => console.error(e));
```

### DIFF
--- a/packages/did-comm/src/types/message-types.ts
+++ b/packages/did-comm/src/types/message-types.ts
@@ -17,6 +17,9 @@ export interface IDIDCommMessage {
   from_prior?: string
   body: any
 }
+export interface IDIDCommOptions {
+  bcc?: string[]
+}
 
 /**
  * Represents different DIDComm v2 message encapsulation
--- a/__tests__/shared/didCommPacking.ts
+++ b/__tests__/shared/didCommPacking.ts
@@ -129,5 +129,74 @@ export default (testContext: {
       expect(unpackedMessage.message).toEqual(message)
       expect(unpackedMessage.metaData).toEqual({ packing: 'authcrypt' })
     })
+
+    it('should pack and unpack message with multiple bcc recipients', async () => {
+      expect.assertions(2)
+
+      const originator = await agent.didManagerCreate({
+        provider: 'did:key',
+      })
+      const beneficiary1 = await agent.didManagerCreate({
+        provider: 'did:key',
+      })
+      const beneficiary2 = await agent.didManagerCreate({
+        provider: 'did:key',
+      })
+
+      const message = {
+        type: 'test',
+        from: originator.did,
+        to: originator.did,
+        id: 'test',
+        body: { hello: 'world' },
+      }
+      const packedMessage = await agent.packDIDCommMessage({
+        packing: 'authcrypt',
+        message,
+        options: { bcc: [beneficiary1.did, beneficiary2.did] },
+      })
+
+      // delete originator's key from local KMS
+      await agent.didManagerDelete({ did: originator.did })
+
+      // bcc'd beneficiaries should be able to decrypt
+      const unpackedMessage = await agent.unpackDIDCommMessage(packedMessage)
+      expect(unpackedMessage.message).toEqual(message)
+      expect(unpackedMessage.metaData).toEqual({ packing: 'authcrypt' })
+    })
+
+    it('should pack and fail unpacking message with multiple bcc recipients', async () => {
+      const originator = await agent.didManagerCreate({
+        provider: 'did:key',
+      })
+      const beneficiary1 = await agent.didManagerCreate({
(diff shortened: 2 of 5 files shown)


mirceanis (changes requested):
Thank you very much for the contribution!

I'd like to propose a slight change that would allow us to maintain closer compatibility with the didcomm spec.

Instead of embedding the bcc property in the message object, pass it in as one of the options to packDidcommMessage.

This way, the DIDCommMessage type remains similar to the spec, and there is no ambiguity about embedding the extra recipients

`IPackDIDCommMessageArgs` can be modified to accept an optional `options` object with a `bcc` property.

The rest of the code would remain the same, I believe, except for loading the list of extra recipients.

inevolin:
> Thank you very much for the contribution!
> 
> I'd like to propose a slight change that would allow us to maintain closer compatibility with the didcomm spec.
> 
> Instead of embedding the bcc property in the message object, pass it in as one of the options to packDidcommMessage.
> 
> This way, the DIDCommMessage type remains similar to the spec, and there is no ambiguity about embedding the extra recipients
> 
> `IPackDIDCommMessageArgs` can be modified to accept an optional `options` object with an optional `bcc` property.
> 
> The rest of the code would remain the same, I believe, except for loading the list of extra recipients.
> 
> It would also be nice to see the new code covered by tests :)

That's even better! Thank you, I'll include tests next as well.

inevolin:
@mirceanis in the new commit:

- refactored bcc to go under new options field
- unit tests (using `did:key` providers because these did not result in errors when resolving them after deletion)
- small syntax improvements

packages/did-comm/src/types/message-types.ts
```diff
@@ -16,6 +16,9 @@ export interface IDIDCommMessage {
   next?: string
   from_prior?: string
   body: any
+  options?: {
```
mirceanis, line 19: This options object should be one of the arguments to `packDidcommMessage`, not part of `IDIDCommMessage`.

Please add it as an optional to `IPackDIDCommMessageArgs` instead

mirceanis (changes requested):
It's still part of the message object, please take another look :)

packages/did-comm/src/types/message-types.ts
inevolin: ahh I see what you meant now, thank you for clarifying, let me correct :)

mirceanis:
Please add it as an optional to `IPackDIDCommMessageArgs`.
I'm not sure why this point is not getting across.

inevolin:
> Please add it as an optional to `IPackDIDCommMessageArgs`. I'm not sure why this point is not getting across.

I wasn't done :) I'll rebase the commits in a bit

inevolin:
Okay should be good now? :D

mirceanis (approved):
excellent!

I'll merge it after the checks pass

Issue #897, opened 2022-05-25
[Veramo/cli] Credential create command does not allow to choose another Subject DID


tientaidev:
**Describe the bug**
Using Veramo cli, when creating a verifiable credential. In input **Subject DID** step, veramo cli does not allow to choose another Subject DID.

**To Reproduce**
Steps to reproduce the behaviour:
1. `veramo credential create`
2. Choose Issuer DID
   
**Expected behaviour**
Cli should allow users to choose another Subject DID

**Details**
![image](https://user-images.githubusercontent.com/43933775/170176036-2b28a5e8-a6a7-4a6a-9c15-5c76cf7c1eee.png)

**Versions (please complete the following information):**
 - Veramo: 3.1.1
 - Node Version: 17.8.0

PR #898, opened 2022-05-26
Add choices when selecting credential Subject in CLI

tientaidev:
Adding choices when select Subject DID instead of always default it to the first identifier. 
This PR fix issues mentioned in #897

### DIFF
--- a/packages/cli/src/credential.ts
+++ b/packages/cli/src/credential.ts
@@ -7,6 +7,8 @@ import * as json5 from 'json5'
 import { readStdin } from './util'
 import { CredentialPayload } from '@veramo/core'
 
+const fuzzy = require('fuzzy')
+
 const credential = program.command('credential').description('W3C Verifiable Credential')
 
 credential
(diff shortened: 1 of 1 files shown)


simonas-notcat (changes requested):
@tientaidev , thanks for your contribution! I would suggest using `type: 'autocomplete'`, so that it's possible to select a DID from a list of known DIDs, but also to manually input any DID. Take a look at how it is done in `sdr` flow:

https://github.com/uport-project/veramo/blob/next/packages/cli/src/sdr.ts#L39

packages/cli/src/credential.ts
```diff
@@ -39,12 +45,24 @@ credential
...
+          const res = fuzzy
+            .filter(input, subjects)
+            .map((el: any) => (typeof el === 'string' ? el : el.original))
+          return res
+        },
+        validate: (val) => {
+          if (val && typeof val !== 'string') {
+            val = val.value
+          }
+          if (!val || !val.startsWith('did:')) {
```
mirceanis, line 61: A subject can be a plain URL, it doesn't have to start with `did:`

It is even OK if the subject is missing, in which case the `credentialSubject.id` property of the credential should be omitted.

mirceanis (changes requested):

The validation is a bit too strict, but otherwise it looks good.

packages/cli/src/credential.ts
mirceanis: if using anything other than a DID causes issues in other places, we can merge this as is and fix the other places later, but please flag it if that is the case

mirceanis (approved):
Looks great,
Thank you!

PR #902, opened 2022-05-29
feat(cli): in explore, allow copy to clipboard the text of identifier or credential or presentation

trentlarson:
Part of the motivation for this is that I could not copy-and-paste with my mouse on the content (which is frustrating... to be able to see it and not copy it!).

Even if I could use my mouse, I figure a quicker way is always helpful. :-)

### DIFF
--- a/packages/cli/src/explore/utils.ts
+++ b/packages/cli/src/explore/utils.ts
@@ -1,4 +1,5 @@
 import { formatDistanceToNow } from 'date-fns'
+import os from 'os'
 
 export function shortDid(did?: string): string {
   if (!did) return ''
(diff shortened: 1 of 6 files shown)


trentlarson:
I did not test this on Windows. I'm comfortable removing that part. (If you believe I must keep it in and test it, that will take quite a bit long.)

nickreynolds (approved):
confirmed that this works as expected on windows. looks good to me!

packages/cli/src/explore/utils.ts
```diff
@@ -12,3 +13,22 @@ export function shortDate(date?: string): string {
...
 }
+
+// return true if copy worked, false otherwise
+export function copyToClipboard(text: string): boolean {
+  /**
+   * There's probably a better way than spawning. We tried and failed at the following:
+   * - screen.copyToClipboard didn't do anything (and is iTerm2-specific, anyway)
+   * - clipboardy wouldn't run due to error "Must use import to load ES Module" (even though we used import)
+   */
+  if (os.platform() === 'darwin') {
```
nickreynolds, line 24: one thing I noticed after my review, this doesn't seem to support linux. imo, it's fine to merge it as is, as long as we create an issue to add linux support (could be a "good first issue")

trentlarson: Good call: [issue logged here](https://github.com/uport-project/veramo/issues/973).

PR #904, opened 2022-05-29
feat(cli): add command to output selected credential / presentation

trentlarson:


trentlarson:
Output is always a single object (ie. not an array), or an error message.

Tested with: empty DB (ie. no credentials or presentations), multiple presentations with the same tag

mirceanis (commented):
Looks good, thank you for the contribution.

Do you have a use case for printing a single credential or presentation?

If you only need to inspect the existing data store, there is the `explore` command, which should allow you to browse multiple other tables too.

packages/cli/src/presentation.ts
```diff
@@ -196,3 +196,56 @@ presentation
...
+            name: 'presentation',
+            choices: list,
+            message: 'Select presentation',
+          },
+        ])
+        selected = answers.presentation
+      }
+
+      if (selected) {
+        console.dir(selected, { depth: 10 })
```
simonas-notcat, line 244: Maybe we should use `console.log(JSON.stringify(selected, null, 2))` instead, so that the putput is a valid json?

mirceanis: Good point

trentlarson: OK, I'll make that change.

trentlarson:
@mirceanis My reason is to extract the VC & VP data, for reference as I build on top of Veramo (and as I build my own tooling). Without these, I had to use sqlite and make queries into the DB to pull out the data. I figure a quicker way is useful.

My edit to the 'explore' may be more interesting to you:
https://github.com/uport-project/veramo/pull/902

However, even with that piece to be able to copy data with an interaction, a command-line utility that can extract the data with a single command seems useful. (It's unfortunate that the output contains yarn data as well.) Multiple ways FTW, IMHO. :-)

packages/cli/src/presentation.ts
trentlarson: I have made that change.

Just note that:
- The `console.dir` command already prints in nicely-formatted JSON with indentation. As a matter-of-fact: it prints in my terminal in such a way that the output is colored, so I think that version is even more fancy and potentially nicer for the user, so it's worth a look to see if you want to change it back.
- The `console.dir` approach is already used in many places in the codebase.

nickreynolds:
I confirmed that this works correctly on windows. However, this work is included in #902 , so I believe this PR should be closed in favor of 902

trentlarson:
> I confirmed that this works correctly on windows. However, this work is included in #902 , so I believe this PR should be closed in favor of 902

D'oh! You're right. My apologies... I tried to keep them separate but made some mistake somewhere. Thanks for pointing it out.

PR #914, opened 2022-06-03
feat(data-store-json): LocalStorageStore

simonas-notcat:


mirceanis (approved):
Looks good.
This is a browser specific feature. Maybe we can also name it to suggest that?

simonas-notcat:
> Looks good.
> This is a browser specific feature. Maybe we can also name it to suggest that?

Maybe `browser-local-storage-store.ts` and `BrowserLocalStorageStore` ?

mirceanis:
> Maybe `browser-local-storage-store.ts` and `BrowserLocalStorageStore` ?

Sure, that sounds good.
I'm also OK with the name you chose too :)

mirceanis (approved):

PR #922, opened 2022-06-08
Improving documentation CI workflows

italobb:
* Moved the scripts for docs generation from `veramo-website` to `veramo`
* The `veramo-website` workflow will run when code is pushed to `main`
* Fixing small issues in the docs generator


scripts/docs-extract.ts
```diff
@@ -0,0 +1,30 @@
+import * as path from 'path'
```
mirceanis, line 1: This script may not be necessary since these same actions are performed during the build by the `extract-api` and `generate-plugin-schema` scripts

.github/workflows/veramo-website.yml
```diff
@@ -0,0 +1,20 @@
+name: Trigger veramo-website build
+on:
+  push:
+    branches:
+      - "main"
+      - "italo/fix-docs"
+permissions: write-all
```
mirceanis, line 7: I think permissions alterations should not be necessary as long as the GH_TOKEN has the required scope.

.github/workflows/veramo-website.yml
```diff
@@ -0,0 +1,20 @@
...
+    permissions:
+      contents: write
+      actions: write 
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          curl -X POST \
+            -H 'Authorization: Bearer ${{ secrets.GH_TOKEN }}' \
+            -H "Accept: application/vnd.github.v3+json" \
+            https://api.github.com/repos/uport-project/veramo-website/actions/workflows/main.yml/dispatches \
```
mirceanis, line 19: It seems safer to call the repository_dispatch event api:
https://docs.github.com/en/rest/repos/repos#create-a-repository-dispatch-event

I believe that if we do that there won't be a need to mess with permissions either.

.github/workflows/build-and-test-on-pr.yml
```diff
@@ -15,13 +15,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-        with:
```
mirceanis, line 18: Why remove this?

.github/workflows/build-and-test-on-pr.yml
```diff
@@ -30,7 +26,9 @@ jobs:
         with:
           path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
           key: ${{ runner.os }}-yarn-2-${{ hashFiles('**/package.json') }}
-
+      - name: Get yarn cache directory path
```
mirceanis, line 29: This new block seems unnecessary, since the cache dir was already used earlier.

.github/workflows/veramo-website.yml
```diff
@@ -0,0 +1,20 @@
+name: Trigger veramo-website build
+on:
+  push:
```
mirceanis, line 3: It would be nice to only run this if the publish workflow succeeds

mirceanis (changes requested):
Generally good, but I have some doubts about the token permissions when dispatching to the other repo.

I think that using the repository_dispatch event may be safer

scripts/docs-extract.ts
italobb: Script removed.

.github/workflows/veramo-website.yml
italobb: Removed.

.github/workflows/veramo-website.yml
italobb: Done.

.github/workflows/build-and-test-on-pr.yml
italobb: Remnants of experiments. Fixed.

.github/workflows/build-and-test-on-pr.yml
italobb: Remnants of experiments. Fixed.

.github/workflows/veramo-website.yml
italobb: Remnants of experiments. Fixed.

.github/workflows/veramo-website.yml
```diff
@@ -0,0 +1,18 @@
...
+      contents: write
+      actions: write 
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          curl -X POST \
+            -H 'Authorization: Bearer ${{ secrets.GH_TOKEN }}' \
+            -H "Accept: application/vnd.github.v3+json" \
+            https://api.github.com/repos/uport-project/veramo-website/actions/workflows/main.yml/dispatches \
+            -d '{"event_type":"deploy_website"}'
```
mirceanis, line 18: I think the repository dispatch event requires another path and they don't mention the Bearer token
https://docs.github.com/en/rest/repos/repos#create-a-repository-dispatch-event

```suggestion
          curl -X POST \
            -H 'Authorization: ${{ secrets.GH_TOKEN }}' \
            -H "Accept: application/vnd.github.v3+json" \
            https://api.github.com/repos/uport-project/veramo-website/dispatches \
            -d '{"event_type":"deploy_website"}'
```

mirceanis (changes requested):
Only one last thing to do, please check the URL for triggering the repository dispatch.
I added a suggestion with what I believe to be the right URL.

Everything else looks great!

mirceanis (approved):
Looks great

PR #924, opened 2022-06-09
feat: kms-web3

### LINKED ISSUE #688: Add support for EthWallets as KMS
IdentityManager has an import function and it would be good to have the ability to import DIDs that have keys that are managed by EthWallets (e.g., MetaMask). This is required to support did:ethr and did:ens using Blockchain Accounts that are only managed in an EthWallet.

Key operations that will be required are likely eth_signTransaction (for updating ENS records), and signTypedData for EthereumEip712Signature2021 to issue VCs etc.

simonas-notcat:
fixes #688


__tests__/shared/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+        const chainId = getChainIdForDidEthr(didDoc.verificationMethod[0])
+        expect(chainId).toEqual(4)
+      }
+    })
+
+    it('should map identifier keys to did doc', async () => {
+      const account = `0xb09b66026ba5909a7cfe99b76875431d2b8d5190`
+      const did = `did:ethr:0x4:${account}`
+      const controllerKeyId = `metamask-${account}`
+      await agent.didManagerImport({
```
mirceanis, line 34: Maybe not to be fixed in this PR, but definitely before the next release, so I'm writing this as a reference.

This import seems crucial to the way the web3-kms will behave.
The key metadata, the `controllerKeyId` and `kid` of the key must all align properly for this to work.

Also, if web3 providers change, there should also be some update to the managed DIDs and keys from `@veramo/did-manager` / `@veramo/key-manager`

mirceanis (changes requested):
Looks good,

There are 2 skipped tests that cover the base functionality of this new package which should not be skipped.

I pointed out some left-over code to be removed and some typos and missing docs.
Also, I made some notes to be used for the next PR(s) addressing this package to make it more generally usable.

packages/kms-web3/src/web3-key-management-system.ts
```diff
@@ -0,0 +1,131 @@
...
+
+type Eip712Payload = {
+  domain: TypedDataDomain
+  types: Record<string, TypedDataField[]>
+  primaryType: string
+  message: Record<string, any>
+}
+
+export class Web3KeyManagementSystem extends AbstractKeyManagementSystem {
+  constructor(private providers: Record<string, Web3Provider>) {
```
mirceanis, line 20: what is the expected key format for the `providers` key-value param?

PR #939, opened 2022-06-14
fix(kms-web3): use ethers _signTypedData

### LINKED ISSUE #938: kms-web3: eth_signTypedData works only in integration tests
Web3 wallets expect `typedData` to be a JSON string. This is different from what `ganache` expects.

simonas-notcat:
Fixes #938


mirceanis (approved):
Looks great

PR #949, opened 2022-06-23
Fix/jwt credential status

### LINKED ISSUE #934: [BUG] `credentialStatus` is not checked for JWT credentials
The new `@veramo/credential-status` plugin is only used during JSON-LD verification.

The status checks, if available, should also be performed for JWT credentials.

When verifying a JWT credential, and if the credential has a `credentialStatus` entry, the `@veramo/credential-w3c` plugin should check if there is a `checkCredentialStatus` function available on the `context.agent` and call it to verify the status.

italobb:
A fix for the bug #934 ([BUG] credentialStatus is not checked for JWT credentials).
It also adds the verification for EIP712 credentials.

Close #934

### DIFF
--- a/__tests__/shared/credentialStatus.ts
+++ b/__tests__/shared/credentialStatus.ts
@@ -1,3 +1,4 @@
+import { CredentialStatus } from 'credential-status'
 import {
   CredentialPayload, IAgentOptions,
   IDataStore,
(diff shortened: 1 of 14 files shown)


__tests__/shared/didCommWithEthrDidFlow.ts
```diff
@@ -96,7 +96,7 @@ export default (testContext: {
       } catch (e) {
         //nop
       }
-      testContext.tearDown()
+      await testContext.tearDown()
```
mirceanis, line 99: great catch!

__tests__/shared/credentialStatus.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    it('should check credentialStatus for JWT credential', async () => {
+      const verifiableCredential = await agent.createVerifiableCredential({
+        credential: rawCredential,
+        proofFormat: 'jwt',
+      })
+      expect(verifiableCredential).toHaveProperty('proof.jwt')
+
+      const verified = await agent.verifyCredential({ credential: verifiableCredential })
+      expect(statusChecked).toBeTruthy()
```
mirceanis, line 67: Related to my earlier comment, this is where you would instead call something like:
```
expect(checkStatus).toHaveReturnedWith(true)
```

mirceanis (changes requested):
It generally looks good.

I made some suggestions about the patterns used in the tests, please take a look and tell me if I'm missing something.

__tests__/shared/credentialStatus.ts
```diff
@@ -0,0 +1,122 @@
...
+  describe('checking Verifiable Credentials status (revocation)', () => {
+    let agent: ConfiguredAgent
+    let identifier: IIdentifier
+    let statusChecked: boolean
+    let rawCredential: CredentialPayload
+
+    beforeAll(async () => {
+      const expectedResult = {}
+      statusChecked = false
+      const checkStatus = jest.fn(async (credential) => {
```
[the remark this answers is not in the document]
italobb: Actually, the `checkStatus` is called inside the `CredentialStatusPlugin` and not in the test itself, that's why I used this pattern instead of a simples verification.

packages/utils/src/credential-utils.ts
```diff
@@ -12,6 +13,7 @@ import {
 import { blake2bHex } from 'blakejs'
 import { decodeJWT } from 'did-jwt'
 import { normalizeCredential, normalizePresentation } from 'did-jwt-vc'
+import { CredentialStatus } from 'credential-status'
```
mirceanis, line 16: This is no longer needed, since the `isRevoked` method is only used in the `credential-w3c` package
```suggestion
```

packages/utils/src/credential-utils.ts
```diff
@@ -1,6 +1,7 @@
 import { asArray, isDefined } from './type-utils'
 import {
   CredentialPayload,
+  IAgentContext,
```
mirceanis, line 4:
```suggestion
```

packages/credential-w3c/src/action-handler.ts
```diff
@@ -16,6 +16,8 @@ import {
   W3CVerifiablePresentation,
 } from '@veramo/core'
 
+import { CredentialStatus } from 'credential-status'
```
mirceanis, line 19: This import should be paired with a declaration of `credential-status` in `package.json` OR we can remove the import entirely and use a local type instead.

This is a very simple data type that can be re-declared internally here, or simply used as `any` type.

packages/credential-eip712/src/agent/CredentialEIP712.ts
```diff
@@ -124,44 +116,191 @@ export class CredentialIssuerEIP712 implements IAgentPlugin {
...
+        }
+      }
+    }
+    else {
+      throw new Error("Recovered Address does not match issuer")
+    }
+
+    if (await isRevoked(credential, context)) {
+      return false;
+    }
```
mirceanis, line 167: This should be removed and only used in credential-w3c
```suggestion
    if (await isRevoked(credential, context)) {
      return false;
    }
```

packages/credential-eip712/src/agent/CredentialEIP712.ts
```diff
@@ -1,32 +1,24 @@
...
-  MANDATORY_CREDENTIAL_CONTEXT,
-  processEntryToArray,
-  resolveDidOrThrow,
-  mapIdentifierKeysToDoc,
-  getChainIdForDidEthr,
+  extractIssuer, getChainIdForDidEthr,
   getEthereumAddress,
-  isDefined
+  isDefined,
+  isRevoked, MANDATORY_CREDENTIAL_CONTEXT, mapIdentifierKeysToDoc, processEntryToArray,
```
mirceanis, line 8:
```suggestion
  MANDATORY_CREDENTIAL_CONTEXT, mapIdentifierKeysToDoc, processEntryToArray,
```

packages/credential-eip712/package.json
```diff
@@ -16,6 +16,7 @@
   "dependencies": {
     "@metamask/eth-sig-util": "^4.0.1",
     "@veramo/core": "^3.1.0",
+    "@veramo/credential-status": "^3.1.0",
```
mirceanis, line 19:
```suggestion
```

Issue #937, opened 2022-06-14
[proposal] Define a credential revocation API


mirceanis:
We need a plugin interface for the credential revocation frontend.

At this time, there are 2 foreseeable actions that this interface should perform:

* One is a call to revoke a credential. Probably this interface will be implemented by an aggregator that can look at a credential that is about to be revoked, inspect the `credentialStatus` entry and then forward the revocation call to the actual credentialStatus driver.

* The other (which may be optional?) is a call that a Credential Issuer might make to build a `credentialStatus` entry.
Many credentialStatus methods work on collections of credentials, so it is likely that when a credential is being created it needs to know its place in the collection (its index, its merkle path, etc)

This task only refers to the interface for such a plugin, and should initially serve as a discussion board.

italobb:
In the picture below, in yellow, we have the current components already implemented when we talk about credential revocation/status in Veramo. In green, a proposal for the components we should build to allow two initial revocation methods: one using local storage (it can implement [Revocation List 2020](https://w3c-ccg.github.io/vc-status-rl-2020/) or a simpler spec) and another one using an EVM network (following the [ethr-status-registry](https://github.com/uport-project/ethr-status-registry)).

![image](https://user-images.githubusercontent.com/1801687/173910062-78340138-9651-472b-b77d-c6e6562602a1.png)

Here is a brief explanation about each component in the diagram:
- `credential-revocation`: implements the interface mentioned by @mirceanis for the revocation and delegates their implementation to plugin specific for each revocation method.
- `revocation-plugin`: an interface for revocation methods plugins.
- `revocation-eth`: a revocation plugin which uses the [ethr-status-registry](https://github.com/uport-project/ethr-status-registry) spec.
- `revocation-local`: a revocation plugin which uses [Revocation List 2020](https://w3c-ccg.github.io/vc-status-rl-2020/) or a simpler method for revocation, storing the status using the agent `data-store`. This component, exceptionally, will also be used by the agent to serve the endpoint used in `credentialStatus` field of the VCs using the revocation method chosen.
- `credential-status-eth`: a library which implements the status verification using [ethr-status-registry](https://github.com/uport-project/ethr-status-registry).
- `revocation-eth`: a library which implements the revocation of VCs using the  [ethr-status-registry](https://github.com/uport-project/ethr-status-registry).

As suggested by @mirceanis, the `credential-revocation` will expose 2 main endpoints:
- `revoke`
    - Params: 
[...]

PR #956, opened 2022-07-01
feat: Adding the interface for credential status manager

italobb:
* Adding the `ICredentialStatus` interface which defines the methods (and respective parameter and results) use to manage credential statuses
* Small refactor in the `credentialStatus` field of VCs in order to reference it in the `ICredentialStatus` interface
* This is the implementation for the proposal describe in https://github.com/uport-project/veramo/issues/937


packages/core/src/types/ICredentialStatus.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * List all the credential status types (methods) available in the current agent instance.
+   */
+  credentialStatusTypes(): Promise<CredentialStatusType[]>
+
+  /**
+   * Returns the status verification of a particular credential.
+   */
+  credentialStatusRead(vc: VerifiableCredential): Promise<CredentialStatusVerification>
```
mirceanis, line 79: I have some doubts if this method should be part of the interface.

I can see how this might make sense internally for StatusList2021 implementation, but not for others.

For example, in EthrStatusRegistry2019, it would return an internal state which might not be consistent with the on-chain state.

It seems that this functionality is already covered by the `ICheckCredentialStatus` interface.

packages/core/src/types/vc-data-model.ts
```diff
@@ -34,6 +34,15 @@ export type CredentialSubject = {
   [x: string]: any
 }
 
+/**
+ * Expresses the credential status type (also referred to as the credential status method). 
+ * It is expected that the value will provide enough information to determine the current status of the credential 
+ * and that machine readable information needs to be retrievable from the URI. 
+ * For example, the object could contain a link to an external document noting whether or not the credential is suspended or revoked.
+ * @see https://www.w3.org/TR/vc-data-model/#status
```
mirceanis, line 42: This explanation about the value does not refer to the `type` property which is a string to identify the status method.
The `value` being described here is the whole `credentialStatus: CredentialStatus` property of a VC.
It looks like the definition from https://www.w3.org/TR/vc-data-model/#status is broken and I raised an issue here: https://github.com/w3c/vc-data-model/issues/891

```suggestion
 *
 * @see https://www.w3.org/TR/vc-data-model/#status
```



BUT why do you think it's necessary to define this wrapper type here in the first place?
Why isn't `string` enough for the type property?

mirceanis (changes requested):
Thank you for adding this!

I added some notes that I'd like you to consider:
* The exported data types should be marked as `@beta` instead of `@public`
* All exported types must have a `@beta` marker and a tsdoc description.
* The `credentialStatusRead` seems un-necessary to me
* The `CredentialStatusType` property could remain a `string` IMO

I also made some suggestions about imports, comments and return types. If you accept the suggestions, the plugin schemas need to be recreated.

packages/core/src/types/vc-data-model.ts
italobb: It gives a more strong semantic for the field to avoid people misunderstanding what that string should represent. It's not necessary, but good to have.

About the explanation, you're right, I missed that.

And thank you for the issue to fix the VC data model status reference.

packages/core/src/types/ICredentialStatus.ts
italobb: I decided to add it here assuming that someone developing a plugin for status creation/update would also implement the verification. But, following the approach we use for VCs, where issuing and verifying is separated and can be implemented in distinct modules, I'll do the same here and remove this method.

packages/core/src/types/ICredentialStatus.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export { DIDDocument, DIDResolutionOptions, DIDResolutionResult } from 'did-resolver'
+import { IPluginMethodMap } from './IAgent'
+import { CredentialStatus, VerifiableCredential } from './vc-data-model'
+
+/**
+ * The arguments expected by a plugin that implements a credential status type/method 
+ * in order to change the status of an issued verifiable credential.
+ * 
+ * Each cretential status type has its own specific parameters according to their spec.
```
mirceanis, line 13:
```suggestion
 * Each credential status type has its own specific parameters according to their spec.
```

packages/core/plugin.schema.json
```diff
@@ -1840,7 +1840,7 @@
               "type": "string"
             },
             "type": {
-              "type": "string"
+              "$ref": "#/components/schemas/CredentialStatusType"
```
mirceanis, line 1843: I think the plugin schemas need to be regenerated.
Please run `yarn build` locally and commit the new results.

packages/core/src/types/ICredentialStatus.ts
```diff
@@ -0,0 +1,80 @@
...
+export interface CredentialStatusGenerateArgs {
+  /**
+   * The credential status type (aka credential status method) to be used in the `cretentialStatus` generation.
+   */
+  type: CredentialStatusType
+}
+
+/**
+ * Credential status manager interface
+ * @public
```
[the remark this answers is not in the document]
mirceanis: We will probably make some adjustments to this interface as we start integrating status methods.
It's best to be consistent with all the interface members and mark this as `@beta` too

mirceanis (changes requested):
Generally looks good.

There are a couple typos and the plugin schema needs to be regenerated.
Please run `yarn build` and recommit

mirceanis:
@italobb I was thinking about some of the data type names we're using here.
We have the `CredentialStatus` data type defined in vc-data-model.ts and we also use the `CredentialStatus` type from the `credential-status` library under the name `CredentialStatusVerification`. It's a bit confusing, IMO.

While this feature is fresh and there are not many backward compatibility issues, what do you think about renaming some of them for a bit more clarity?

I propose naming them like this:
`CredentialStatus` -> `CredentialStatusCoordinates` for the property that gets embedded in the VC
`CredentialStatusVerification` -> `CredentialStatus` for the current status of the credential

What do you think?
(It doesn't have to be part of this PR, but I commented here because it's related to the subject)

mirceanis:
> @mirceanis About the data type names. After fixing the past issues, the dependency with `credential-status` was removed and we are only referencing the `CredentialStatus` from `vc-data-model.ts`. But I agree this confusion can come back in the future and we should open another issue to fix that. I suggest `CredentialStatusReference` instead of `...Coordinates`.

mirceanis (approved):
looks good.

There's still an issue with the plugin schema for the `@veramo/credential-status` package which seems to be caused by the `CredentialStatusVerification` renamed import.

It will probably be fixed along with #981

PR #958, opened 2022-07-04
chore(ci): add PR template

mirceanis:
This adds a Github PR template.

Hopefully, if all these steps are taken before a PR is opened, we will have an easier time reviewing and accepting contributions.

Does the template make sense?

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/bugfix.md
+++ b/.github/PULL_REQUEST_TEMPLATE/bugfix.md
@@ -1,0 +1,26 @@
+---
+name: Bug fix
+about: Create a PR to improve this project
+title: ''
+labels: triage
+assignees: ''
+
+---
+
+**What issue is this PR fixing**
+Example:
+`closes #123`
+`fixes #456`
+
+**What is being changed**
+A clear description of what this PR brings.
+
+**Quality**
+* [ ] I successfully ran `yarn`, `yarn build`, `yarn test`, `yarn test:browser` locally
+* [ ] I allow my PR to be updated
+* [ ] I added unit tests
+* [ ] I added integration tests
+* [ ] I did not add automated tests because _________, and I am aware that a PR without tests will likely get rejected.
+
+**Details**
+If applicable, add screen captures, error messages or stack traces to help explain your problem.
--- a/.github/PULL_REQUEST_TEMPLATE/bugfix.md
+++ b/.github/PULL_REQUEST_TEMPLATE/bugfix.md
@@ -1,6 +1,6 @@
 ---
 name: Bug fix
-about: Create a PR to improve this project
+about: Create a PR to fix an issue
 title: ''
 labels: triage
 assignees: ''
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE/bugfix.md
```diff
@@ -0,0 +1,26 @@
...
+
+**What is being changed**
+A clear description of what this PR brings.
+
+**Quality**
+* [ ] I successfully ran `yarn`, `yarn build`, `yarn test`, `yarn test:browser` locally
+* [ ] I allow my PR to be updated
+* [ ] I added unit tests
+* [ ] I added integration tests
+* [ ] I did not add automated tests because _________, and I am aware that a PR without tests will likely get rejected.
```
italobb, line 23: All the items in this check list are very important in order to have the PR accepted. So, I think this item is not necessary.

.github/PULL_REQUEST_TEMPLATE/bugfix.md
```diff
@@ -0,0 +1,26 @@
...
+
+**What issue is this PR fixing**
+Example:
+`closes #123`
+`fixes #456`
+
+**What is being changed**
+A clear description of what this PR brings.
+
+**Quality**
```
italobb, line 18: All the items in the checklist below are already checked in our CI/CD flows. If the build doesn't work locally, it'll also fail in CI, and if it doesn't has unit or integration tests we also can add rules in the coverage tool in order to stop the PR. I understand that the checklist helps all to avoid low quality PRs to be created, but I personally don't see that as a problem, because the person will work to fix the issues in the PR detected during the CI verifications. I'd remove this checklist and improve the CI verifications instead.

italobb: @mirceanis  Could you explain your point of view on this, please?

mirceanis: It happens a lot that folks post a PR but don't check back to see if the CI scripts failed or if anyone is interested in reviewing the PR.

This checklist should help with that, since folks might be more aware of what will fail

.github/PULL_REQUEST_TEMPLATE/bugfix.md
```diff
@@ -0,0 +1,26 @@
...
+* [ ] I successfully ran `yarn`, `yarn build`, `yarn test`, `yarn test:browser` locally
+* [ ] I allow my PR to be updated
```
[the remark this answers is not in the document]
mirceanis: This is for the following scenario:
A PR is great, except for some minor things like typos and such, which could be corrected by github suggestions.

If the PR doesn't allow updates, this makes us wait until the PR creator accepts the suggestions and pushes an update.

.github/PULL_REQUEST_TEMPLATE/bugfix.md
mirceanis: I guess I wasn't thinking about it as a checklist, but more like "check all the boxes that apply".

Good catch, I should make that more clear.

I want to keep this last item because I want to nudge folks into adding tests for everything and understanding in advance why their PRs might not be accepted.

Plus, it takes precious energy and time to ask for tests when there are none.

.github/PULL_REQUEST_TEMPLATE/bugfix.md
mirceanis: There is one more thing.

For first time contributors we have to manually run the workflows so it makes more sense to have these things already checked locally before waiting for CI/CD to do its thing.

italobb (approved):

PR #967, opened 2022-07-22
Fix typo in error message for `addKey`

nick-verida:
`addKey` error message is currently incorrect.


mirceanis (approved):
Good catch, thanks!

Issue #971, opened 2022-07-29
[proposal] Add `didManagerUpdate` method to `IDIDManager` & implement `IDIDManagerUpdateArgs` to support full payload update operation


Eengineer1:
By adding this method DIDManager supports whole did payload updates for providers that intentionally follow this workflow, namely `cheqd`.

In the future, this can easily allow arbitrary data update by calculating diffs client-side as well.
In the current scope, diffs are calculated ledger side, submitting whole body payloads in each request.

Eengineer1:
Working on a PR with a proposed solution.

Issue #960, opened 2022-07-07
[proposal] Align IDIDManager interface with DID Registration spec


mirceanis:
### Discussed in https://github.com/uport-project/veramo/discussions/959

<div type='discussions-op-text'>

<sup>Originally posted by **peacekeeper** July  7, 2022</sup>
Hello,

At DIF we've been working on a specification called "DID Registration", which defines inputs and outputs of the DID create/update/deactivate operations, see here:
https://identity.foundation/did-registration/

Just like DID Resolution, these are abstract functions which can be implemented in various ways, e.g. as a local SDK, remote HTTP service, etc. There is also the Universal Registrar implementation of this at DIF: https://github.com/decentralized-identity/universal-registrar

The DID Registration specification seems to have some overlap with Veramo's [IDIDManager](https://veramo.io/docs/api/core.ididmanager/) interface. Perhaps it would be useful to align your interface with the specification, or conversely, incorporate ideas from your interface into the specification.

Markus
</div>

Issue #948, opened 2022-06-23
[proposal] add `metadata` property to `IIdentifier`


mirceanis:
**Is your feature request related to a problem? Please describe.**
DIDs are detailed publicly by their DID document, but an agent might need to store additional information about a DID.

For example, linking multiple DIDs to a common known entity privately, or referring to different recovery or update methods that do not get reflected in the DID document.

It is already possible to add additional private information to DIDs by modifying the database schema, but that can be a bit cumbersome, so a generic API for this should be created.

**Describe the solution you'd like**
The `IIdentifier` interface should have a `metadata` property which would be a freeform JSON object that can be set and modified through the DID manager using a new plugin method.

A new plugin method on `IDIDManager` should be added, `didManagerUpdateMetadata(): Promise<IIdentifier>`, with these arguments:

```typescript
type DIDMetadata = {
  [x: Exclude<string, keyof IIdentifier>]: any;
}

type IDIDManagerUpdateMetadataArgs = {
  did: string, // the DID to be updated
  meta: DIDMetadata, // the new metadata
  replace?: boolean = false // flag indicating if the existing metadata object is replaced completely or just the properties defined above
}
```

The existing keys of the `IIdentifier` interface MUST NOT be used as keys in the `DIDMetadata`.
`didManagerUpdateMetadata` will throw an error if these keys are being used.

**Describe alternatives you've considered**
* Some other properties of the `IIdentifier` can be overloaded and used as generic storage using a stringified JSON object (`alias` and `controllerKeyId`). It would be up to DID providers to use these properties appropriately.
* The `keys` and `services` arrays can also contain data unrelated to the DID document, but updating these is usually not a simple internal operation so some special logic might have to be added to the respective DID providers to handle these purely local updates. It is possible to update these by reimporting the `IIdentifier`, but that also feels hacky.

**Additional context**
Please see [this discussion in our discord server](https://discord.com/channels/878293684620234752/935484187639705610/989472608435208232).
You may need to join first using [this invite link](https://discord.gg/aDDFhnTA5N).

PR #974, opened 2022-08-01
feat(did-manager-update): Added `IDIDManager` `didManagerUpdate` method for supported providers

Eengineer1:
Linked to Issue #971 .

### DIFF
--- a/packages/core/src/types/IDIDManager.ts
+++ b/packages/core/src/types/IDIDManager.ts
@@ -106,7 +106,7 @@ export interface IDIDManagerGetOrCreateArgs {
   /**
    * Identifier alias. Can be used to reference an object in an external system
    */
-  alias: string
+  alias?: string
 
   /**
    * Optional. Identifier provider
@@ -121,6 +121,28 @@ export interface IDIDManagerGetOrCreateArgs {
   /**
    * Optional. Identifier provider specific options
    */
+  options?: object
+}
+
+export interface IDIDManagerUpdateArgs {
+  /**
+   * Required. DID
+   */
+  did: string
+
+  /**
+   * Identifier alias. Can be used to reference an object in an external system
+   */
+  alias?: string
+
+  /**
+   * Optional. Key Management System
+   */
+  kms?: string
+
+  /**
+    * Optional. Identifier provider specific options
+    */
   options?: object
 }
 
@@ -299,6 +321,8 @@ export interface IDIDManager extends IPluginMethodMap {
     args: IDIDManagerGetOrCreateArgs,
     context: IAgentContext<IKeyManager>,
   ): Promise<IIdentifier>
+
+  didManagerUpdate(args: IDIDManagerUpdateArgs, context: IAgentContext<IKeyManager>): Promise<IIdentifier>
 
   /**
    * Imports identifier
--- a/packages/core/src/types/IDIDManager.ts
+++ b/packages/core/src/types/IDIDManager.ts
@@ -1,3 +1,4 @@
+import { DIDDocument } from 'did-resolver'
 import { IPluginMethodMap, IAgentContext } from './IAgent'
 import { IIdentifier, IService, IKey, MinimalImportableIdentifier } from './IIdentifier'
 import { IKeyManager } from './IKeyManager'
@@ -131,19 +132,14 @@ export interface IDIDManagerUpdateArgs {
   did: string
 
   /**
-   * Identifier alias. Can be used to reference an object in an external system
-   */
-  alias?: string
-
-  /**
-   * Optional. Key Management System
-   */
-  kms?: string
-
-  /**
-    * Optional. Identifier provider specific options
+    * Identifier provider specific options
     */
-  options?: object
+  options: {
+    /**
+     * Partial pre-constructed DIDDocument
+     */
+    document: Partial<DIDDocument>
+  }
 }
 
 /**
--- a/packages/core/src/types/IDIDManager.ts
+++ b/packages/core/src/types/IDIDManager.ts
@@ -107,7 +107,7 @@ export interface IDIDManagerGetOrCreateArgs {
   /**
    * Identifier alias. Can be used to reference an object in an external system
    */
-  alias?: string
+  alias: string
 
   /**
    * Optional. Identifier provider
@@ -132,13 +132,15 @@ export interface IDIDManagerUpdateArgs {
   did: string
 
   /**
+   * Required
+   */
+  document: Partial<DIDDocument>
+
+  /**
     * Identifier provider specific options
     */
-  options: {
-    /**
-     * Partial pre-constructed DIDDocument
-     */
-    document: Partial<DIDDocument>
+  options?: {
+    [x: string]: any
   }
 }
 
--- a/packages/did-manager/src/abstract-identifier-provider.ts
+++ b/packages/did-manager/src/abstract-identifier-provider.ts
@@ -10,8 +10,8 @@ export abstract class AbstractIdentifierProvider {
     context: IAgentContext<IKeyManager>,
   ): Promise<Omit<IIdentifier, 'provider'>>
 
-  abstract updateIdentifier(
-    args: { did: string, options: { document: Partial<DIDDocument> } },
+  abstract updateIdentifier?(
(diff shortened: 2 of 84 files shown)


packages/core/src/types/IDIDManager.ts
```diff
@@ -124,6 +124,28 @@ export interface IDIDManagerGetOrCreateArgs {
   options?: object
 }
 
+export interface IDIDManagerUpdateArgs {
```
mirceanis, line 127: The shape of this interface suggests that the update being done is mostly, if not only, to the locally known `IIdentifier` metadata.
If this is meant to update the DIDDocument, I think it should be more suggestive of that, by adding a parameter of type `DIDDocument` to represent the new DID doc.

I imagine something like this would be more suited for full document updates:
```ts
export interface IDIDManagerUpdateDocumentArgs {
  did: string,
  document: DIDDocument, // this could also be `Partial<DIDDocument>` // or a new type that can also link to local keys (read below)
  options?: Record<string, any> // these options get forwarded to the DID provider
}
```

---

If the new DID document contains new keys, there's also the question of whether the agent should be managing these keys or not.

If the agent already knows the keys, the agent could figure out which keys match the update and update the local `IIdentifier.keys` array accordingly.

Alternatively, the `verificationMethod` array from the `document` parameter could contain private key information, in which case these keys would also be imported by the `KeyManager`. In this later case, the `DIDDocument` data type would have to be replaced by an extended version that includes veramo specific key information. However, this would introduce some complexity, so it would be better if the user would first import the keys directly to the key manager and then call this update.

What do you think?

packages/core/src/types/IDIDManager.ts
```diff
@@ -124,6 +124,28 @@ export interface IDIDManagerGetOrCreateArgs {
...
+  /**
+   * Required. DID
+   */
+  did: string
+
+  /**
+   * Identifier alias. Can be used to reference an object in an external system
+   */
+  alias?: string
```
mirceanis, line 136: Is the `alias` supposed to be updated, or is it being used to find some data related to this update?

I think it would be a good idea to separate local updates from remote (ledger) updates.
See also #948 for the proposal to update local metadata (`alias` would eventually fall under the metadata category, even if it is treated as a top level property now)

packages/core/src/types/IDIDManager.ts
```diff
@@ -124,6 +124,28 @@ export interface IDIDManagerGetOrCreateArgs {
...
+
+  /**
+   * Optional. Key Management System
+   */
+  kms?: string
```
mirceanis, line 141: I think this is not really needed here since the plugin methods get access to `context` which should allow them to call `IKeyManager` methods.
Specifying a particular `kms` here suggests that the `DIDManager` plugin should also know how to work with key management systems directly, which is not the case.

I suppose this would be needed if during the update, this agent should also be importing keys into a particular KMS. I prefer to keep these 2 actions separate.

mirceanis (changes requested):
Thank you for contributing this!

I have some questions and suggestions that I'd like to clarify before going forward with this. Please take a look and share your thoughts.


This is also related to #960 and #948

packages/core/src/types/IDIDManager.ts
Eengineer1: The `alias` property can be skipped, as we are not passing the `provider` property, rather than relying on `did` as the predicate.
Initially, the rationale was to pass `alias` to be accessible to the provider. In that case, it can also be passed as metadata to the `options` object, which in turn is passed to the provider.
I will remove it from a top level arg.

packages/core/src/types/IDIDManager.ts
Eengineer1: The rationale is to always pass a raw DIDDoc along with key-pairs.

The workflow:

- Import keys, if needed for the specific DIDDoc (might already be managed). Handled by `didManagerImport` for an identifier or `keyManagerImport` for just keys.
- Update on ledger, handling validation logic ledger-side
- Return the identifier implementing `IIdentifier`

The provider is going to handle the ledger-specific args format and key-pairing in signature generation.
The `DIDManager`'s `didUpdateMethod` only takes care of passing the needed information to the provider handling the external update and updates the local identifier implementing `IIdentifier`.

packages/core/src/types/IDIDManager.ts
Eengineer1: Good point!
This can be handled by `Import` methods, either `didManagerImport` or `keyManagerImport`.

packages/core/src/types/IDIDManager.ts
Eengineer1: Also, about the DIDDoc, I'd prefer to pass it in the options, as they are being directly passed to the provider as well.

e.g. 
```typescript
export interface IDIDManagerUpdateArgs {
    did: string
    options?: {
        document: DIDDocument
    }
}
```

@mirceanis What do you think?

packages/did-manager/src/id-manager.ts
```diff
@@ -133,13 +135,46 @@ export class DIDManager implements IAgentPlugin {
...
+      const identifierProvider = this.getProvider(identifier.provider)
+      const updatedIdentifier = await identifierProvider.updateIdentifier(
+        { did, options },
+        context,
+      )
+      await this.store.import(updatedIdentifier)
+      return updatedIdentifier
+    } catch {
+      throw Error('Provider does not support updateIdentifier')
+    }
```
mirceanis, line 175: It's probably better to let the lower level errors bubble up here so that the developer gets the more precise failure reason, instead of catching and re-throwing as a generic error message.

If the intent is to throw when this method is not implemented, then it might be better to first check for the type before calling it:
 
```ts
      if (typeof identifierProvider?.updateIdentifier !== 'function') {
        throw new Error(`not_supported: ${identifier?.provider} provider does not implement full document updates`)
      }
      const updatedIdentifier = await identifierProvider.updateIdentifier(
      // ...
```

```suggestion
```

packages/did-manager/src/abstract-identifier-provider.ts
```diff
@@ -10,6 +10,11 @@ export abstract class AbstractIdentifierProvider {
     context: IAgentContext<IKeyManager>,
   ): Promise<Omit<IIdentifier, 'provider'>>
 
+  abstract updateIdentifier(
```
mirceanis, line 13: This method can be made optional:

```suggestion
  abstract updateIdentifier?(
```

mirceanis (changes requested):
Excellent progress!

Please check my suggestions and let me know what you think

packages/did-manager/src/id-manager.ts
Eengineer1: Great point. Included.

packages/core/src/types/IDIDManager.ts
```diff
@@ -106,7 +107,7 @@ export interface IDIDManagerGetOrCreateArgs {
   /**
    * Identifier alias. Can be used to reference an object in an external system
    */
-  alias: string
+  alias?: string
```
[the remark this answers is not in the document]
Eengineer1: This is reverted back to the original. Might have been a typo.

packages/core/src/types/IDIDManager.ts
```diff
@@ -124,6 +125,23 @@ export interface IDIDManagerGetOrCreateArgs {
...
+export interface IDIDManagerUpdateArgs {
+  /**
+   * Required. DID
+   */
+  did: string
+
+  /**
+    * Identifier provider specific options
+    */
+  options: {
```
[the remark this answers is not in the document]
Eengineer1: Nesting the DIDDoc was mostly to cover for possible compatibility issues.
Switching back to top level arg for simplicity purposes as well.

About `options`, reverted to optional and a proposed impl following your suggestion for extensibility.

packages/did-manager/src/abstract-identifier-provider.ts
Eengineer1: Agreed. Included & removed extra throwing from other providers.

mirceanis (approved):
Wonderful!

Thank you very much for the contribution.

PR #982, opened 2022-08-09
feat(credential-status): rename plugin interfaces and methods

### LINKED ISSUE #981: [chore] clarify credential-status method names
Based on a discussion from a previous issue(https://github.com/uport-project/veramo/pull/956#issuecomment-1206160383):

---

We have the `CredentialStatus` data type defined in `vc-data-model.ts` and we also use the `CredentialStatus` type from the `credential-status` library under the name `CredentialStatusVerification`. It's a bit confusing, IMO.

While this feature is fresh and there are not many backward compatibility issues, what do you think about renaming some of them for a bit more clarity?

I propose naming them like this:
`CredentialStatus` -> `CredentialStatusCoordinates` for the property that gets embedded in the VC
`CredentialStatusVerification` -> `CredentialStatus` for the current status of the credential

@italobb proposed `CredentialStatusReference` instead of `CredentialStatusCoordinates`

This should be done before the v4 release.

mirceanis:

fixes #981

## What is being changed

I renamed the credential status interfaces to `ICredentialStatusVerifier` and `ICredentialStatusManager` and created a `ICredentialStatus` plugin interface to represent both.
I moved these interfaces to `@veramo/core` and updated the plugin schema generated.
The schema generation had some issues because of some inline docs so I fixed the inline tsdocs that were creating issues.

There is an extra change in this PR that was not related to #981:
The `ICredentialStatusVerifier.checkCredentialStatus` method `didDoc` parameter was renamed to `didDocumentOverride` and is now optional.
If it is not provided, the `IResolver` plugin is called to resolve the issuer DID document. I also added a test for this.

## Quality
Check all that apply:
* [X] I added unit tests.


packages/core/src/types/vc-data-model.ts
```diff
@@ -44,14 +44,29 @@ export type CredentialSubject = {
...
+ *
+ * Implementations MUST populate the `revoked` boolean property, but they can return additional metadata that is
+ * method specific.
+ *
+ * @see {@link credential-status#CredentialStatus | CredentialStatus}
+ *
+ * @beta This API may change without a BREAKING CHANGE notice.
+ */
+export type CredentialStatus = {
+  revoked?: boolean
```
italobb, line 64: If implementations MUST populate `revoked`, we shouldn't mark it as optional, right?

italobb (changes requested):
Just one issue found related to the `revoked` field, in the `CredentialStatus` class, which is still an optional field but in the comments we say it MUST be filled by the implementations.

packages/core/src/types/vc-data-model.ts
italobb:
```suggestion
  revoked: boolean
```

mirceanis: You are right. It has to be non-optional.

mirceanis: @italobb I made the `revoked` property required and added a check for the `StatusMethod` returned value to look for the `revoked` property.
An error is thrown if the property is missing, and I adjusted the tests to include this.

italobb: Great! Thanks.

italobb (approved):

PR #988, opened 2022-08-10
feat: add support for serviceEndpoint property as defined in latest DID Spec

nickreynolds:

The latest DID Spec (https://www.w3.org/TR/did-core/), recommended 19 July 2022, defines the `serviceEndpoint` property (a sub-property of `service`, a property of the DID Document) as:

 >A [string](https://infra.spec.whatwg.org/#string) that conforms to the rules of [[RFC3986](https://www.w3.org/TR/did-core/#bib-rfc3986)] for [URIs](https://www.w3.org/TR/did-core/#dfn-uri), a [map](https://infra.spec.whatwg.org/#string), or a [set](https://infra.spec.whatwg.org/#ordered-set) composed of a one or more [strings](https://infra.spec.whatwg.org/#string) that conform to the rules of [[RFC3986](https://www.w3.org/TR/did-core/#bib-rfc3986)] for [URIs](https://www.w3.org/TR/did-core/#dfn-uri) and/or [maps](https://infra.spec.whatwg.org/#string).
 
 Previously, Veramo (in several packages) assumed that the `serviceEndpoint` property was simply a `string`.

## What is being changed
This PR adds support for `serviceEndpoint` properties that are defined as `maps`, or `sets` of `strings` or `maps`. Integration tests were also added to cover the changes.

## Quality
Check all that apply:
* [ ] I added unit tests.

## Details
This PR also includes upgrades to several dependencies required for this functionality:
- `"did-resolver": "^4.0.0"`
- `"ethr-did-resolver": "^6.0.2"`
- `"web-did-resolver": "^2.0.20"`
- `"did-jwt": "^6.3.0"`
- `"did-jwt-vc": "^2.1.14"`
- `"credential-status": "^2.0.5"`


italobb (approved):

packages/data-store/src/identifier/did-store.ts
```diff
@@ -138,7 +149,7 @@ export class DIDStore extends AbstractDIDStore {
       const service = new Service()
       service.id = argsService.id
       service.type = argsService.type
-      service.serviceEndpoint = argsService.serviceEndpoint
+      service.serviceEndpoint = JSON.stringify(argsService.serviceEndpoint)
```
mirceanis, line 152: There should be a type check here, because if `argsService.serviceEndpoint` is already a string, it will get serialized as `"serviceEndpoint"` instead of `serviceEndpoint` (quotes added).

packages/did-comm/src/transports/transports.ts
```diff
@@ -100,17 +100,43 @@ export class DIDCommHttpTransport extends AbstractDIDCommTransport {
...
 
   /** {@inheritdoc AbstractDIDCommTransport.send} */
   async send(service: any, message: string): Promise<IDIDCommTransportResult> {
+    let serviceEndpointUrl = ''
+    if (typeof service.serviceEndpoint === 'string') {
+      serviceEndpointUrl = service.serviceEndpoint
+    } else if (service.serviceEndpoint.uri) {
+      serviceEndpointUrl = service.serviceEndpoint.uri
+    } else if (service.serviceEndpoint.length > 0) {
+      if (typeof service.serviceEndpoint[0] === 'string') {
```
mirceanis, line 131: somehow I'm sure picking just the first element will come back to byte us later, but I'm not sure if this should be treated differently at this point

packages/did-provider-ethr/src/ethr-did-provider.ts
```diff
@@ -314,7 +314,7 @@ export class EthrDIDProvider extends AbstractIdentifierProvider {
 
     debug('ethrDid.setAttribute %o', { attrName, attrValue, ttl, gasLimit })
 
-    const txHash = await ethrDid.setAttribute(attrName, attrValue, ttl, undefined, {
+    const txHash = await ethrDid.setAttribute(attrName, JSON.stringify(attrValue), ttl, undefined, {
```
mirceanis, line 317: please type-check this to avoid double quoting strings

packages/did-provider-ethr/src/ethr-did-provider.ts
```diff
@@ -360,7 +360,7 @@ export class EthrDIDProvider extends AbstractIdentifierProvider {
     const gasLimit = args.options?.gasLimit || this.gas || DEFAULT_GAS_LIMIT
 
     debug('ethrDid.revokeAttribute', { attrName, attrValue, gasLimit })
-    const txHash = await ethrDid.revokeAttribute(attrName, attrValue, undefined, {
+    const txHash = await ethrDid.revokeAttribute(attrName, JSON.stringify(attrValue), undefined, {
```
mirceanis, line 363: this also needs a type-check

packages/did-comm/src/didcomm.ts
```diff
@@ -603,7 +603,9 @@ export class DIDComm implements IAgentPlugin {
         }
 
         debug('Sending to %s', serviceEndpoint)
-        const res = await fetch(serviceEndpoint, {
+        const endpointUri = (typeof serviceEndpoint === 'string' || serviceEndpoint instanceof String) ? serviceEndpoint : serviceEndpoint.uri
```
mirceanis, line 606: is this OR really needed here?
` || serviceEndpoint instanceof String`

It's not a request to change, just trying to learn in case there is some corner case that I'm not aware of.

mirceanis (changes requested):
Very nice, thanks for doing the leg-work from low level libs up to here, and kudos for adding all the test cases!

There are some cases of possibly over-serialized strings, I think. Please take a look.

packages/did-comm/src/didcomm.ts
nickreynolds: no you're probably right. I had seen this pattern somewhere, but we don't use it throughout our project and I think it's unnecessary

packages/did-comm/src/transports/transports.ts
nickreynolds: yeah, there's an existing `TODO` about "better strategy for selecting the transport" but it's unclear what that strategy would be

mirceanis: Let's leave it as is and solve this when it's an issue for someone and the expectations are more clear

mirceanis (approved):
Looks great!

Issue #375, opened 2021-02-08
Generalized Verification API in Veramo Core


rado0x54:
As an Integrator of Veramo I would like to be able to use a single generalized Verification API that automatically verifies a various set of input data according to a given policy input.

Acceptance Criteria:
- There should be a single Verification API within Veramo core (Proposal: `agent.verify({ payload: {...}, policy: {...} })`)
- The fist step of the verification API is the identification of the payload format and determines what dynamic module is able to verify the given Payload.
- The generalized verification API allows for the registration of Modules to support new payloads.
- In the initial Implementation VC-JWT and VC-JSON-LD Credentials are supported with respective VC-JWT and VC-JSON-LD Verification Modules.
- Furthermore it should be able to Verify Verifiable Presentation Representations of JWT / JSON-LD Credentials
- The Verification API and the Verification Modules take an optional policy file that is able to influence the verification outcome.
- Each VC Verification Module is implemented independently, but generally it needs to verify the following.
a) Schema / Data format. It needs to verify that the given payload matches the referenced Schema or Context
b) Signature Verification. Verify that a Issuer Signature matches the given Credential and Issuer.
c) Credential Status: Verify that the Credential Status of the VC is still valid:
- Policy files can model the following:
a) Override certain assumption: Policy may allow to accept revoked credentials or self-signed Signatures.
b) Tighten Validation conditions: Further restrict conditions of credential verification (e.g. similar to passing a PE), e.g. where certain values are expected. (Credential issued by specific issuer A)
c) Trust Frameworks
d) ... TBD
- Fixed Error Types for Verification Errors
a)
b)
c)
d)
- Policy Design, Module Integration and Payload Identification will be design as part of this ticket.

rado0x54:
Figure out if there should be a Verifiable Presentation vs Verifiable Credential separation of interfaces. (in addition to a universal layer).

italobb:
I would like to reinforce the importance of this issue hoping it can be prioritized.
The usage of the [DIF "well-known DID configuration" spec](https://identity.foundation/specs/did-configuration/) with the purpose of implementing discoverability and interoperability requires such support from Veramo.

mirceanis:
It is easier to start with a more granular approach to verification instead of cramming everything into a thin top-level API.
For this reason, I think it's best to start with different methods for verifying credentials and presentations.
There is already a big gap between LD-proofs and JWT that indicates that we can go even more granular if necessary.

[...]

PR #990, opened 2022-08-12
Added policies to verifyPresentation in credential-w3c module

### LINKED ISSUE #954: Error Handling in Veramo Credential-w3c package.
We are working heavily with the Veramo credential-w3c package and whenever we try to verify a presentation which is created with a broken credential for example with invalid issuance date or expiration date, we do not receive the error for why it failed and only receive a boolean stating false.

We would like to see the reason for why the verification failed as well. To find out what is broken in the credential or the verifiable presentation.

**Additional context**
Code snippet where the error needs to be returned.
```
try { 
    const verification = await verifyPresentationJWT(jwt, resolver, {
      challenge: args.challenge,
      domain: args.domain,
      audience,
    })
    return true
} catch (e: any) {
    //TODO: return a more detailed reason for failure
    return false
}
```

daniyal-khalil:

partially fixes #954

## What is being changed
Adding the `IVerifyPresentionResponse` interface in the credential-w3c module to be able to implement better error handling, the new interface allows the errors to be communicated through an Object containing fields, `verified`, `details`, `errorCode`.
- `verified` defines the state of the verification process, whether it failed or is passed.
- `details` contains details regarding the failure for the verification process.
- `errorCode` can be used to pass machine readable errorCodes from the lower level modules.

## Quality
Check all that apply:
* [X] I added unit tests.

## Details
Was unable to find specific verification tests in the `action-handler.test.ts` in the credentials-w3c module so created another test suite like in credentialLD module for issue-verify-flow.


packages/credential-w3c/src/action-handler.ts
```diff
@@ -215,13 +215,52 @@ export interface IVerifyPresentationArgs {
...
+  exp?: boolean
+}
+
+/**
+ * Encapsulates the response object to verifyPresentation method after verifying a
+ * {@link https://www.w3.org/TR/vc-data-model/#presentations | W3C Verifiable Presentation}
+ *
+ * @public
+ */
+ export interface IVerifyPresentationResponse {
```
mirceanis, line 245: Let's make this object extensible to allow various other details to be included in the result:
```suggestion
 export interface IVerifyPresentationResponse {
   [x:string]: any
   
```

packages/credential-w3c/src/action-handler.ts
```diff
@@ -573,11 +612,17 @@ export class CredentialIssuer implements IAgentPlugin {
...
+        return { verified: true }
       } catch (e: any) {
-        //TODO: return a more detailed reason for failure
-        return false
+        const errorCodes = ['invalid_jwt','invalid_config', 'invalid_signature', 'not_supported', 'no_suitable_keys', 'resolver_error']
+        if (errorCodes.some(errorCode => e.message.includes(errorCode))) {
+          return { verified: false, details: e.message }
+        }
+        else {
+          throw e
```
mirceanis, line 624: I really like this separation of concerns between legitimate errors and exceptional situations.

However, I'm not yet 100% sure we should be introducing an extra branch, though.
I'm concerned about the usability here where developers have to handle 3 situations: `valid response`, `invalid response` and `thrown error`.
This is not a critique of your approach (which I've also described in the past), but rather me venting my uncertainty :)

packages/credential-w3c/src/__tests__/issue-verify-flow.test.ts
```diff
@@ -0,0 +1,155 @@
+import {
```
mirceanis, line 1: I always appreciate new tests being added!

mirceanis (changes requested):
Kudos for attacking this problem from all fronts (including did-jwt)!

This is a good start but introduces some quirks too:
* the policies are JWT specific
* the new response only deals with JWT presentations, other formats are not handled which makes the API unclear.
* It would also be nice to have a similar approach to VCs too

If you don't mind, I'll mark this PR as draft until all the pieces align and we are able to bring in the policy support from did-jwt.

packages/credential-w3c/src/action-handler.ts
```diff
@@ -215,13 +215,52 @@ export interface IVerifyPresentationArgs {
...
+    */
   [x: string]: any
 }
 
+export interface VerifyPresentationPolicies {
+  now?: number
+  nbf?: boolean
+  iat?: boolean
+  exp?: boolean
+}
```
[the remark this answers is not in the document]
daniyal-khalil: Added this support

tmp/local-database-0.26140.json
```diff
@@ -0,0 +1 @@
...8f4999ddbc3fa30678910ce8348defbcbf73e260da3c6c615acb39ab6dc448c24d6568d61f098":{"type":"Secp256k1","kid":"04a67d3ec015713102efbeea960c9a413bbafd1e101000bcc89ad8f4999ddbc3fa30678910ce8348defbcbf73e260da3c6c615acb39ab6dc448c24d6568d61f098","publicKeyHex":"04a67d3ec015713102efbeea960c9a413bbafd1e101000bcc89ad8f4999ddbc3fa30678910ce8348defbcbf73e260da3c6c615acb39ab6dc448c24d6568d61f098","meta":{"algorithms":["ES256K","ES256K-R","eth_signTransaction","eth_signTypedData","eth_signMessage"],"did":"did:ethr:rinkeby:0x02a67d3ec015713102efbeea960c9a413bbafd1e101000bcc89ad8f4999ddbc3fa"},"kms":"local"}},"privateKeys":{"04a67d3ec015713102efbeea960c9a413bbafd1e101000bcc89ad8f4999ddbc3fa30678910ce8348defbcbf73e260da3c6c615acb39ab6dc448c24d6568d61f098":{"alias":"04a67d3ec015713102efbeea960c9a413bbafd1e101000bcc89ad8f4999ddbc3fa30678910ce8348defbcbf73e260da3c6c615acb39ab6dc448c24d6568d61f098","type":"Secp256k1","privateKeyHex":"e07a959a493016b6ee6c0dfb2cf2aa25aa5578c287b62c06ac2879406188aa08eee8d7a9fa42d612687fe690b9faa0e978f6d89573ec1c6f42c9be88b86e767e26866c7942863dfc1beb73cbfbc07c0eef626e9b45525f59e0960e449207aa3b3cd4ad46b3a0bf62"}},"credentials":{},"claims":{},"presentations":{},"messages":{}}
```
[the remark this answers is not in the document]
daniyal-khalil: Sorry for this, removed it.

packages/credential-w3c/src/action-handler.ts
```diff
@@ -215,13 +215,52 @@ export interface IVerifyPresentationArgs {
...
+  /**
+   * This value is used to transmit the result of verification.
+   */
+  verified: boolean
+
+
+  /**
+   * Optional string only to be used to convey the response message from errors.
+   */
+  details?: string
```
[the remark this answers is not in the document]
daniyal-khalil: Abstract away the error object, and created a new IError Type in veramo/core which can be used for error handling in other verification modules as well.

packages/credential-w3c/src/action-handler.ts
```diff
@@ -215,13 +215,52 @@ export interface IVerifyPresentationArgs {
...
    * Other options can be specified for verification.
    * They will be forwarded to the lower level modules. that performt the checks
-   */
+    */
   [x: string]: any
 }
 
+export interface VerifyPresentationPolicies {
+  now?: number
+  nbf?: boolean
```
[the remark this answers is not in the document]
daniyal-khalil: Changed them around to full forms, but they are not the disable one, in did-jwt merge request to disable these policies you have to make them false. So they do the default behaviour if `true` or `undefined`.

packages/credential-w3c/src/action-handler.ts
daniyal-khalil: Updated this as well.

packages/credential-w3c/src/action-handler.ts
daniyal-khalil: Understandable, how do you propose we approach this issue?

daniyal-khalil:
@mirceanis I was wondering whether you'd prefer if I create a `IVerifyResponse` as a type in Veramo core which will have a general `verified` and `error` fields. This can be used generally in all Verification APIs in separate modules.

tmp/local-database-0.81314.json
```diff
@@ -0,0 +1 @@
...9Gkq3hB2MayAW0-fHx1TY4ToKwknYaRL_UzzybgUtBp2NfyEuvLnBRK9oq0w"}}]}},"messages":{"d1e91f722c3963ec7e240dc092e59bf8d408f71f7ab3126019ffd6616c4d9b2e7510c0efbf248436a5814ad5a5634a5596d4c4b663d371926ef60652a6e4435f":{"raw":"eyJ0eXAiOiJKV1QiLCJhbGciOiJFUzI1NksifQ.eyJpYXQiOjE2NTg5OTgzMDIsInR5cGUiOiJzZHIiLCJ0YWciOiJzZHItb25lIiwiY2xhaW1zIjpbeyJyZWFzb24iOiJXZSBuZWVkIGl0IiwiY2xhaW1UeXBlIjoibmFtZSIsImVzc2VudGlhbCI6dHJ1ZX1dLCJpc3MiOiJkaWQ6ZXRocjpyaW5rZWJ5OjB4MDM0NTA2YzYxNDZhZGU0MTg1YmJiMzA4OTZlNWM0ZmVhN2U3Yzg4Y2E1ZDIxMDI0ODA4ZDk2ZjI0ZjQ1ZDQ3NDQzIn0.kec_DXu-5BH5I4z3_d26Mn5hIIfSR5XFuKq4o5sITUN81CZPtolax0Uuz8RdrpFVC7X4DmMoER8LIBFU0B7NKQ","metaData":[{"type":"JWT","value":"ES256K"}],"data":{"iat":1658998302,"type":"sdr","tag":"sdr-one","claims":[{"reason":"We need it","claimType":"name","essential":true}],"iss":"did:ethr:rinkeby:0x034506c6146ade4185bbb30896e5c4fea7e7c88ca5d21024808d96f24f45d47443"},"id":"d1e91f722c3963ec7e240dc092e59bf8d408f71f7ab3126019ffd6616c4d9b2e7510c0efbf248436a5814ad5a5634a5596d4c4b663d371926ef60652a6e4435f","type":"sdr","from":"did:ethr:rinkeby:0x034506c6146ade4185bbb30896e5c4fea7e7c88ca5d21024808d96f24f45d47443","threadId":"sdr-one","createdAt":"2022-07-28T08:51:42.000Z"}}}
```
mirceanis, line 1: tmp files should not be added here

packages/credential-w3c/src/action-handler.ts
```diff
@@ -573,11 +619,17 @@ export class CredentialIssuer implements IAgentPlugin {
           challenge: args.challenge,
           domain: args.domain,
           audience,
+          policies: args.policies
```
mirceanis, line 622: I think policies should be translated here to match the JWT specific ones expected by did-jwt

packages/credential-w3c/src/action-handler.ts
mirceanis: I suppose the simple answer would be to not re-throw an error in the exceptional case, but to return an object just like for the legitimate failures.
The difference would be that for the legitimate failures we can also identify the cause and populate the  `errorCode` property of the error.

packages/credential-w3c/src/action-handler.ts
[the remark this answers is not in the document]
mirceanis: that's great, but don't forget to translate them so that did-jwt understands the overrides

packages/core/src/types/IError.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Error interface
+ * @public
+ */
+
+export interface IError {
```
mirceanis, line 6:
```suggestion
/**
 * An error object, which can contain a code.
 * @beta
 */
export interface IError {
```

packages/credential-w3c/src/action-handler.ts
```diff
@@ -215,6 +216,51 @@ export interface IVerifyPresentationArgs {
...
+  policies?: VerifyPresentationPolicies
+
+  /**
+   * Other options can be specified for verification.
+   * They will be forwarded to the lower level modules. that performt the checks
+    */
+  [x: string]: any
+}
+
+export interface VerifyPresentationPolicies {
```
mirceanis, line 233:
```suggestion
/**
 * These optional settings can be used to override some of the default checks that are performed on
 * Presentations during verification.
 *
 * @beta
 */
export interface VerifyPresentationPolicies {
```

mirceanis (changes requested):
There are still some quirks left to be fixed:
* the policies are forwarded as-is to the lower levels
* The new result type is only returned for JWT presentations. Non-JWT presentation types should also be covered to have a usable API.
If you are willing to create different PR(s) to cover the non-JWT types, please say so, but otherwise we'll have to fix it in this PR.

I also made some comment suggestions and a compatibility suggestion for the new IError type

mirceanis:
> @mirceanis I was wondering whether you'd prefer if I create a `IVerifyResponse` as a type in Veramo core which will have a general `verified` and `error` fields. This can be used generally in all Verification APIs in separate modules.

Yes, that sounds like a good idea, although `IVerifyResult` fits the action better, since these are not actually request/response pairs.

tmp/local-database-0.81314.json
daniyal-khalil: sorry again, removed these.

packages/credential-w3c/src/action-handler.ts
daniyal-khalil: Yep, understood.

daniyal-khalil:
I think even if I proceed to create separate PRs for the `credential-ld` and `credential-eip712` modules. I would still need to refactor the `verifyPresentationLD` to support the current PR because the `else` clause responds with the LD verification. Which I believe is incompatible with the current `IVerifyResult` Response.

daniyal-khalil:
If we align on the approach, then I think it would not take long to implement similar logic for all the `verify*` methods.

mirceanis (approved):
looks good

More PRs are needed to align the verification API, but this will do for starters.

This also relates to #375

Issue #994, opened 2022-08-15
Replace "revoked" by "verified" in credential status verification


italobb:
There are some JSON-LD libraries that work with credential-status which, instead of an object with a `revoked` property, adopted an object with a `verified` property, similar to the result of a VC verification. It allows developers:
- to issue credentials that can also be suspended, instead of only a  revoked.
- to issue credentials in an invalid state and later turned valid.
- to more easily switch to Veramo without writing adapters for them.

PR #996, opened 2022-08-19
feat(credential-w3c): align verification API between formats

### LINKED ISSUE #375: Generalized Verification API in Veramo Core

### LINKED ISSUE #935: [proposal] Selective verification properties for verifyCredential/Presentation
The boolean output of these methods may not be sufficient to convey the various ways in which credentials may be seen as invalid. Also, some verifications may not be required in all cases (for example, checking if a credential was valid in the past).

Credentials are legitimately invalid if:
* formatted incorrectly
* expired or not valid yet
* signature invalid (according to the respective algorithm)
* signature does not match any of the issuer keys
* credentialStatus check resulted in revoked

Presentations, additionally must pass domain and challenge checks.

For these legitimate cases, the verification result must be `false`, but the returning output could contain additional information for the failure.

---

Credentials or presentations cannot have a definitive verification status if:

* Signature suite not installed.
* `DID` method not installed on resolver
* issuer `DID`/`kid` resolution failed
* credentialStatus method not installed
* credentialStatus check failed
* @context processing error for LD credentials

It is arguably OK to throw exceptions in these situations, but the error messages should guide developers.

### LINKED ISSUE #954: Error Handling in Veramo Credential-w3c package.

### LINKED ISSUE #989: How do you issue credential with custom Context?
I am trying to issue credentials using a custom context and I can't figure out how to get this to work. It seems veramo is using @digitalcredentials/vc under the hood and that specifies that if you want a custom context, you need to set a documentLoader:

https://github.com/digitalcredentials/vc#custom-documentLoader

As for the custom context, I'm actually just taking the existing example one used everywhere from:
https://www.w3.org/2018/credentials/examples/v1
and serving it from my own server, with the correct `application/ld-json` content type. And that fails with 
```
jsonld.InvalidUrl: Dereferencing a URL did not result in a valid JSON-LD object. Possible causes are an inaccessible URL perhaps due to a same-origin policy (ensure the server uses CORS if you are using client-side JavaScript), too many redirects, a non-JSON response, or more than one HTTP Link Header was provided for a remote context.
    at ContextResolver._fetchContext (/project/test/node_modules/jsonld/lib/ContextResolver.js:173:13)
    at async ContextResolver._resolveRemoteContext (/project/test/node_modules/jsonld/lib/ContextResolver.js:117:34)
    at async ContextResolver.resolve (/project/test/node_modules/jsonld/lib/ContextResolver.js:50:22)
    at async api.process (/project/test/node_modules/jsonld/lib/context.js:65:20)
    at async api.expand (/project/test/node_modules/jsonld/lib/expand.js:211:17)
    at async Function.jsonld.expand (/project/test/node_modules/jsonld/lib/jsonld.js:325:18)
    at async Function.jsonld.toRDF (/project/test/node_modules/jsonld/lib/jsonld.js:671:16)
    at async Function.jsonld.normalize.jsonld.canonize (/project/test/node_modules/jsonld/lib/jsonld.js:580:19) {
  details: {
    code: 'loading remote context failed',
    url: 'http://localhost:3000/api/view/339709073577476172',
    cause: Error: Document loader unable to load URL "http://localhost:3000/api/view/339709073577476172".
        at documentLoader (/project/test/node_modules/@digitalcredentials/vc/lib/documentLoader.js:18:9)
        at Object.<anonymous> (/project/test/node_modules/@digitalcredentials/jsonld-signatures/lib/documentLoader.js:48:12)
        at LdCredentialModule.<anonymous> (/project/test/node_modules/@veramo/credential-ld/src/ld-credential-module.ts:91:17)
        at Generator.next (<anonymous>)
        at /project/test/node_modules/@veramo/credential-ld/build/ld-credential-module.js:31:71
        at new Promise (<anonymous>)
        at __awaiter (/project/test/node_modules/@veramo/credential-ld/build/ld-credential-module.js:27:12)
        at /project/test/node_modules/@veramo/credential-ld/src/ld-credential-module.ts:37:54
        at /project/test/node_modules/@digitalcredentials/jsonld-signatures/lib/documentLoader.js:48:12
        at /project/test/node_modules/@digitalcredentials/jsonld-signatures/lib/documentLoader.js:48:12
```

Am I doing something wrong? How are others using custom contexts?

mirceanis:

fixes #935
fixes #954
fixes #375
fixes #989

## What is being changed
* Verification override policies are now defined for Credentials as well, and, where possible, applied to JSON-LD VC/VP too (see #935).
* Verification of credentials and presentations results in an object with a `verified` boolean property and an optional `error` object with a `message` and `errorCode`. Other properties can be added by the lower level libraries performing the verification (see #954).
* `verify*()` methods still throw errors, but only in case of exceptional situations, when the agent config does not support that kind of verification or when there is a problem with the arguments.
  * I'm still not sure if this is the right approach or if we should be catching these and creating response objects with error messages. **Please share your opinion if you have one.** If needed, this change can be added in another PR.
* VC/VP creation and verification now allows extra options to be used and forwarded to lower level libraries (fixes #989)

## Quality
Check all that apply:
* [ ] I added unit tests. (policies are applied by lower level libs which have their own unit tests for this)

## Details
Not all policies are applicable for non-JWT VC/VP.
For EIP712 VC/VP, none of the override policies are applicable yet.

This PR is a continuation of the work started by @daniyal-khalil

### DIFF
--- a/packages/credential-w3c/src/action-handler.ts
+++ b/packages/credential-w3c/src/action-handler.ts
@@ -113,6 +113,15 @@ export interface ICreateVerifiablePresentationArgs {
   keyRef?: string
 
   /**
+   * When dealing with JSON-LD you also MUST provide the proper contexts.
+   * Set this to `true` ONLY if you want the '@context' URLs to be fetched in case they are not preloaded.
+   * The context definitions SHOULD rather be provided at startup instead of being fetched.
+   *
+   * @default false
+   */
+  fetchRemoteContexts?: boolean
+
+  /**
    * Any other options that can be forwarded to the lower level libraries
    */
   [x: string]: any
@@ -164,6 +173,15 @@ export interface ICreateVerifiableCredentialArgs {
   keyRef?: string
 
   /**
+   * When dealing with JSON-LD you also MUST provide the proper contexts.
+   * Set this to `true` ONLY if you want the '@context' URLs to be fetched in case they are not preloaded.
+   * The context definitions SHOULD rather be provided at startup instead of being fetched.
+   *
+   * @default false
+   */
+  fetchRemoteContexts?: boolean
+
+  /**
    * Any other options that can be forwarded to the lower level libraries
    */
   [x: string]: any
@@ -188,7 +206,7 @@ export interface IVerifyCredentialArgs {
 
   /**
    * When dealing with JSON-LD you also MUST provide the proper contexts.
-   * Set this to `true` ONLY if you want the '@context' URLs to be fetched in case they are not pre-loaded.
+   * Set this to `true` ONLY if you want the '@context' URLs to be fetched in case they are not preloaded.
    * The context definitions SHOULD rather be provided at startup instead of being fetched.
    *
    * @default false
--- a/packages/credential-w3c/src/action-handler.ts
+++ b/packages/credential-w3c/src/action-handler.ts
@@ -275,7 +275,7 @@ export interface VerificationPolicies {
   /**
    * policy to skip the revocation check (credentialStatus) when set to `false`
    */
-  credentialStatus: boolean
+  credentialStatus?: boolean
 
   /**
    * Other options can be specified for verification.
@@ -443,7 +443,7 @@ export class CredentialIssuer implements IAgentPlugin {
           verifiablePresentation = await context.agent.createVerifiablePresentationLD(args)
         } else {
           throw new Error(
-            'invalid_configuration: your agent does not seem to have ICredentialIssuerLD plugin installed',
+            'invalid_setup: your agent does not seem to have ICredentialIssuerLD plugin installed',
           )
         }
       } else if (args.proofFormat === 'EthereumEip712Signature2021') {
@@ -451,7 +451,7 @@ export class CredentialIssuer implements IAgentPlugin {
           verifiablePresentation = await context.agent.createVerifiablePresentationEIP712(args)
         } else {
           throw new Error(
-            'invalid_configuration: your agent does not seem to have ICredentialIssuerEIP712 plugin installed',
+            'invalid_setup: your agent does not seem to have ICredentialIssuerEIP712 plugin installed',
           )
         }
       } else {
@@ -519,7 +519,7 @@ export class CredentialIssuer implements IAgentPlugin {
           verifiableCredential = await context.agent.createVerifiableCredentialLD(args as any)
         } else {
           throw new Error(
-            'invalid_configuration: your agent does not seem to have ICredentialIssuerLD plugin installed',
+            'invalid_setup: your agent does not seem to have ICredentialIssuerLD plugin installed',
           )
         }
       } else if (args.proofFormat === 'EthereumEip712Signature2021') {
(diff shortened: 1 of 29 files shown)


packages/credential-w3c/src/action-handler.ts
```diff
@@ -672,21 +805,22 @@ function wrapSigner(
...
 }
 
-async function isRevoked(credential: VerifiableCredential, context: IAgentContext<any>): Promise<boolean> {
+async function isRevoked(credential: VerifiableCredential, context: IContext): Promise<boolean> {
   if (!credential.credentialStatus) return false
 
   if (typeof context.agent.checkCredentialStatus === 'function') {
     const status = await context.agent.checkCredentialStatus({ credential })
-    return status?.revoked == true
+    return status?.revoked == true || status?.verified === false
```
mirceanis, line 820: relates to #994

@italobb I'm not sure if this is enough to fix it

simonas-notcat (approved):

nickreynolds (approved):

PR #1001, opened 2022-09-07
feat(credential-w3c): add ICredentialPlugin interface in core package

### LINKED ISSUE #941: [proposal] split `ICredentialIssuer` verification methods to a new `ICredentialVerifier` interface
Naming is hard, but what we have now is an interface and a plugin implementation for both issuers and verifiers.
It would be better to separate them, or rename the existing interface.

Both options are viable if we also create an interface that inherits from both.

mirceanis:

closes #941

## What is being changed

* The `ICredentialIssuer` interface was moved to the `@veramo/core` package, but is reexported by the `@veramo/credential-w3c` package for compatibility.
* `verify*()` methods from the `ICredentialIssuer` interface were isolated into a `ICredentialVerifier` interface (also in `@veramo/core`)
* Created a `ICredentialPlugin` interface to represent both issuer and verifier
* Renamed `CredentialIssuer` plugin to `CredentialPlugin` (The old name is still exported, but deprecated)

## Quality
Check all that apply:
* [X] I updated unit tests.
* [X] I updated integration tests.


italobb (approved):
The refactoring seems good.

PR #1022, opened 2022-10-09
feat(credential-status-router): Added basic router implementation & verifier/ manager encapsulation

Eengineer1:
Ref on architecture on Veramo's Discord #credential-status.
This is a router that handles top-level routing logic for different status methods in one agent, extending the `AbstractStatusMethod` abstract class each.

Base methods inherit from existing `ICredentialStatusVerifier` & `ICredentialStatusManager`.

Preparing a thorough coverage test case version, but the structure should suffice for `credential-status-list-2021` tests.


mirceanis (commented):
Thanks for posting this!

I need help understanding some of the methods defined.
I'd really like to maintain the separation of concerns between Verifier and Manager

Issue #583, opened 2021-06-21
[proposal] Add interface for changing the DID controller


mirceanis:
Veramo should have a common API for changing a DID controller. As a prerequisite, the `AbstractIdentifierProvider` class should require implementers to also implement a `changeController` method.
A proposal for this new method signature is this:
```ts
abstract changeController(
    args: { identifier: IIdentifier; newController: string, options?: any },
    context: IAgentContext<IKeyManager & IDIDManager>,
  ): Promise<string>
```

Of all the `did-provider-*` packages that are currently implemented in this repository, this method would only have a concrete implementation in `did-provider-ethr`, in which case the `newController` would be an ethereum address and the returned `string` would be the transaction hash.
The `options` parameter can be used to provide transaction value overrides (`gasPrice`, `gasLimit`, `nonce`, etc..).

It is expected that the `context.agent.keyManagerSign()` method would be called to sign the transaction and that the `controllerKeyId` of the `identifier` would be updated.

### Originally discussed in https://github.com/uport-project/veramo/discussions/574

<div type='discussions-op-text'>

<sup>Originally posted by **atz3n** June 17, 2021</sup>
Hi all 👋,

Is there an interface for changing the DID controller? I couldn't find such a function within the `AbstractIdentifierProvider` class. If not, maybe it is a good idea to define one to have a convenient API for this task.

Cheers,
Attila</div>

lleifermann:
@mirceanis Lets pick this topic up again. I really like the idea of the proposed interface you had there. Thinking about did:ethr and other methods one thing comes to my mind. 

Right now did:ethr does not support adding multiple owners, while other methods may support this (did:web may just add multiple keys to the controller field) - so the Interface `changeOwner` always suggest that it replaces some existing value. This may conflict with other implementations.

How about we add 3 abstract methods to the Abstract that an IdentifierProvider may leverage. 

- `changeController` - For replacing an owner (can be implemented in did:ethr)
- `addController` - For adding a controller did and keeping the existing one (cant be implemented in did:ethr)
- `removeController` - For removing a controller did (also cant be implemented in ethr) 

ethr-did would only implement `changeController` then. The only reason i see the add and remove methods is to keep consistent with the current methods, since 'change' suggest replacement why may not be wanted by other methods. Then we are in the territory of passing options again :smile:

lleifermann:
@mirceanis Any update on this? :smile:

mirceanis:
> @mirceanis Any update on this? 😄

[...]

PR #1031, opened 2022-10-12
feat: add support for did:ethr signed/meta transactions

lleifermann:
This PR adds support for sending transactions on behalf of another did:ethr.  

This is archived via updates to the underlying dependencies https://github.com/uport-project/ethr-did & https://github.com/decentralized-identity/ethr-did-resolver. 

To keep abstraction intact, this behavior may be triggered by passing a `metaIdentifier` to a didManager action which is picked up by the `did-provider-ethr`. If a metaIdentifier is present the underlying controller is passed this identifier as a signer which then signs and pays for the transactions. 

The methods `addKey` & `addService` now detect this presence as well and prepare the transaction differently then. 
Overall we tried to keep changes to other plugins as minimal as possible. 

Some todo'ish items: 

- Right now only addKey and addService have meta capabilities. `changeOwner` and `addDelegate` are NOT implemented yet. @mirceanis - Not sure how to tackle this with the current AbstractIdentifierProvider. We would have to put these actions into the `updateIdentifier` method. But with the current abstraction the did-provider-ethr would receive a didDocument there. Not sure how to proceed here with a good approach. 
We could change the abstract signature of `updateIdentifier` to something more easily hand-able (like actions being passed down or something). 

- We had to introduce a new 'algorithm' to the kms-local provider. All other existing options for creating eth signatures either modify the input data, or hash it again. We just need clean digest with the `r, s and v` params easily accessible afterwards. 

- The tests for the did-provider-ethr changes will fail as i am awaiting an update to ethr-did which fixes an encoding bug we found.

- After some testing and fiddling around we also noticed that if this package does not pass any options down to did-ethr some defaults for transactions will be chosen, which could result in the transaction failing. Gaslimit and Gasprice are always overridden as defaults. AFAIK if etherjs is not passed anything it will make a best guess for determining these parameters.

### DIFF
--- a/__tests__/shared/ethrDidFlowSigned.ts
+++ b/__tests__/shared/ethrDidFlowSigned.ts
@@ -1,0 +1,139 @@
+// noinspection ES6PreferShortImport
+
+import {
+  IAgentOptions,
+  IDIDManager,
+  IIdentifier,
+  IKeyManager,
+  IMessageHandler,
+  IResolver,
+  TAgent,
+} from '../../packages/core/src'
+import {IDIDComm} from '../../packages/did-comm/src'
+// @ts-ignore
+import express from 'express'
+import {Server} from 'http'
+
+type ConfiguredAgent = TAgent<IDIDManager & IKeyManager & IResolver & IDIDComm & IMessageHandler>
+
+export async function sleep(milliseconds: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, milliseconds))
+}
+
+export default (testContext: {
+  getAgent: () => ConfiguredAgent
+  setup: (options?: IAgentOptions) => Promise<boolean>
+  tearDown: () => Promise<boolean>
+}) => {
+  describe('did ethr controller signed interactions', () => {
+    let agent: ConfiguredAgent
+
+    let alice: IIdentifier
+    let bob: IIdentifier
+
+    let didCommEndpointServer: Server
+    let listeningPort = Math.round(Math.random() * 32000 + 2048)
+
+    beforeEach(async () => {
+      await testContext.setup()
+      agent = testContext.getAgent()
+
+      alice = await agent.didManagerImport({
+        controllerKeyId: 'alice-controller-key',
+        did: 'did:ethr:ganache:0x0279be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798',
+        provider: 'did:ethr:ganache',
+        alias: 'alice-did-ethr',
+        keys: [
+          {
+            privateKeyHex: '0000000000000000000000000000000000000000000000000000000000000001',
+            kms: 'local',
+            type: 'Secp256k1',
+            kid: 'alice-controller-key',
+          },
+        ],
+      })
+
+      bob = await agent.didManagerImport({
+        controllerKeyId: 'bob-controller-key',
+        did: 'did:ethr:ganache:0x02c6047f9441ed7d6d3045406e95c07cd85c778e4b8cef3ca7abac09b95c709ee5',
+        provider: 'did:ethr:ganache',
+        alias: 'bob-did-ethr',
+        keys: [
+          {
+            privateKeyHex: '0000000000000000000000000000000000000000000000000000000000000002',
+            kms: 'local',
+            type: 'Secp256k1',
+            kid: 'bob-controller-key',
+          },
+        ],
+      })
+    })
+
+    afterAll(async () => {
+      await testContext.tearDown()
+    })
+
+    it('should add dummy service to alice did with bob sending the tx', async () => {
+      const result = await agent.didManagerAddService({
+        did: alice.did,
+        service: {
+          id: 'localhost-useless-endpoint',
+          type: 'DIDComm',
+          serviceEndpoint: `http://localhost:${listeningPort}/foobar`,
+          description: 'this endpoint will be removed',
+        }, options: {
+          metaIdentifier: bob.did
+        }
+      })
+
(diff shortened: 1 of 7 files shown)


packages/kms-local/src/key-management-system.ts
```diff
@@ -130,6 +130,8 @@ export class KeyManagementSystem extends AbstractKeyManagementSystem {
         return await this.eth_signMessage(managedKey.privateKeyHex, data)
       } else if (['eth_signTypedData', 'EthereumEip712Signature2021'].includes(algorithm)) {
         return await this.eth_signTypedData(managedKey.privateKeyHex, data)
+      } else if (['eth_rawSign'].includes(algorithm)) {
```
mirceanis, line 133: This KMS should also advertise this algorithm on the `meta.algorithms` property of the keys it creates.
Please see the `asManagedKeyInfo` method below.

At some point we will want to have finer control over the KMS chosen for particular operations based on the key capabilities (`meta.algorithms`), so I think it's worth keeping these 2 in sync.

mirceanis (changes requested):
Thanks for posting this!
Some tiny changes required regarding `eth_rawSign` and the unused `ethers` dependency.

---

On a broader note, from what I can see, the expectation here is that the Veramo instance is managing both the DID controller key and the transaction signing key. This does reduce some friction with regards to funding every controller key, but the real power of meta-transactions and of our modular key-management/did-management architecture would come from splitting these 2 actions (of signing the change and of sending the transaction) completely.

This would mean that `didManagerAddKey()` with a "meta-transaction" flag, would not actually send a transaction, but only perform the DID-controller signing. The result of this operation would then be sent to a transaction-signing service, which would be a separate thing.

This service could, of course, be implemented by the same Veramo instance, but it could also be a third party service.

I'm not requesting this change be added here, only flagging this as a future improvement.
Let me know what you think.

lleifermann:
@mirceanis i removed the ethers dependency and moved the ethr_rawSign into it's own method and is also now advertised in the meta.algorithm array for Secp256k1 keys.

Regarding the signature flow we wanted to keep it as straight forward as possible. But i see your point in separating these things. 

Whats your take on the missing `changeOwner` and `addDelegate` functions? Should we place those in the `updateIdentifier` function? AFAIK this would require a small change in the Abstract IdentifierProvider as it right now passes passes a DidDocument down. We figured a key/value map of possible `UpdateActions` would fit better as the concrete implementation of the provider can then expose them properly. (did-provider-ethr providing changeOwnerAction and addDelegateAction for example). Other providers then may provide other actions which fit in the `updateIdentifier` action.

With the current implementation we would need to do some delta comparison i believe (current did doc vs new did doc = wanted change)

mirceanis:
> @mirceanis i removed the ethers dependency and moved the ethr_rawSign into it's own method and is also now advertised in the meta.algorithm array for Secp256k1 keys.

That's great, thanks!

> Regarding the signature flow we wanted to keep it as straight forward as possible. But i see your point in separating these things.

It makes sense. What you did here is very valuable too, even if the full power of meta-tx is not used yet.

> Whats your take on the missing `changeOwner` and `addDelegate` functions? Should we place those in the `updateIdentifier` function? AFAIK this would require a small change in the Abstract IdentifierProvider as it right now passes passes a DidDocument down. We figured a key/value map of possible `UpdateActions` would fit better as the concrete implementation of the provider can then expose them properly. (did-provider-ethr providing changeOwnerAction and addDelegateAction for example). Other providers then may provide other actions which fit in the `updateIdentifier` action.
> 
> With the current implementation we would need to do some delta comparison i believe (current did doc vs new did doc = wanted change)

This is a tricky problem, because a change in `AbstractIdentifierProvider` would have to make sense for most DID methods (or at least more than the one we're dealing with now). My gut reaction is that custom control logic, that only makes sense to a certain DID method should be built as its own plugin and not try to morph the `DIDManager` API, but DID controller updates are probably common?
We started discussing this issue in #583 (wow, long time ago!), so we can continue the chat about this there.

__tests__/shared/ethrDidFlowSigned.ts
```diff
@@ -82,13 +82,19 @@ export default (testContext: {
...
         }, options: {
-          metaIdentifierKeyId: bob.did
+          metaIdentifierKeyId: bob.controllerKeyId
         }
       })
 
       const resolution = await agent.resolveDid({ didUrl: alice.did })
 
-      expect(resolution?.didDocument?.service?.[0].serviceEndpoint).toEqual(`http://localhost:${listeningPort}/foobar`)
+      expect(resolution.didDocument).toEqual(expect.objectContaining({
```
mirceanis, line 91: TIL about arrayContaining/objectContaining :)
This looks very clean

mirceanis (approved):

lleifermann:
Only thing missing here is now the failing import of the `ethr-did-controller`. Once the downstream bundling of `ethr-did` and `ethr-did-controller` is fixed i can include the version here. Then everything should be good to go as far as i can tell. 

For the addDelegate and changeOwner stuff i would propose a solution in #583

__tests__/shared/keyManager.ts
```diff
@@ -493,75 +506,80 @@ export default (testContext: {
...
-          ]
+              name: 'type',
+              type: 'string[]',
+            },
+          ],
         },
-        "message": {
-          "issuer": {
-            "id": "did:fake:123"
+        message: {
```
mirceanis, line 564: @lleifermann looks like there's some code duplication here causing test failures

lleifermann: Fixed it. My prettier messed something up i believe.

PR #1035, opened 2022-10-19
didcomm libp2p fixes

nickreynolds:


packages/core/src/types/IKeyManager.ts
```diff
@@ -214,6 +219,9 @@ export interface IKeyManager extends IPluginMethodMap {
    * Returns an existing key
    */
   keyManagerGet(args: IKeyManagerGetArgs): Promise<IKey>
+  keyManagerGetWhere(args: IKeyManagerGetWhereArgs): Promise<ManagedKeyInfo>
```
mirceanis, line 222: I'm surprised we didn't already have a function like this.

We definitely need a way to find keys by
* type
* kms
* the DIDs they are associated with
* algorithms
* metadata matches (this implies unstructured JSON queries)

packages/did-comm/src/didcomm.ts
```diff
@@ -194,7 +194,19 @@ export class DIDComm implements IAgentPlugin {
     }
 
     // obtain sender signing key(s) from authentication section
-    const senderKeys = await mapIdentifierKeysToDoc(managedSender, 'authentication', context)
+
+    // can't sign did-comm message using web3 kms
+    let senderKeys: _ExtendedIKey[] = (await mapIdentifierKeysToDoc(managedSender, 'verificationMethod', context)) //.filter(k => k.kms !== 'web3')
```
mirceanis, line 199: please also flag this as a workaround, as technically picking from `keyAgreement` is correct;
it's just very very picky, and annoying, but from a security perspective, it does reflect the intent of the DID controller for the purpose of keys.

Feel free to ignore this comment if this PR is only experimental

packages/libp2p-client/src/client.ts
```diff
@@ -75,11 +71,16 @@ export class AgentLibp2pClient implements IAgentPlugin {
               // console.log("msg of source: ", msg)
               message = message + (msg.toString().replace('\n',''))
             }
-            context?.agent.handleMessage({ raw: message })
+            context?.agent.handleMessage({ raw: message, save: true })
```
mirceanis, line 74: we're trying to deprecate the `save` param and encourage users to deliberately save messages that make sense for them after handling(parsing) by calling methods of the data-store

mirceanis (commented):
These workarounds make sense, but it also raises questions about the ease of use of our current did-comm flows, especially for did:ethr.

I added some inline comments, not knowing what the final purpose of this PR is.

packages/did-comm/src/didcomm.ts
nickreynolds: Can you expand on this a bit? It would seem like there's a bug somewhere in did-ethr if `keyAgreement` is correct. Why would the Managed DID not include the new key in keyAgreement after I've added it? Am I adding the key incorrectly?

I'd like to find a more permanent solution, as this workaround causes some tests to fail

mirceanis: I meant that using keys from their correct verification relationship should be the default.

It's usually [not safe to use the same key for multiple purposes](https://github.com/christianlundkvist/blog/blob/master/2020_05_26_secp256k1_twist_attacks/secp256k1_twist_attacks.md). If you add a secp256k1 key to the `assertionMethod` relationship, it shouldn't be also used for `keyAgreement`, for example.

Picking keys from the `verificationMethod` pool instead of from the `keyAgreement` or `authentication` sections respectively is a workaround.

We do have a bug in the `did-provider-ethr` in that we don't allow the user to decide where to add new keys, and just assume based on the key type.

To simplify some of the work needed to get did-comm working using two did:ethr, I also created this issue: https://github.com/decentralized-identity/ethr-did-resolver/issues/174

PR #1039, opened 2022-10-23
Support for NIST Secp256r1 keys and ES256 signatures

nklomp:

Add support for NIST Secp256r1 keys and ES256 signatures

## What is being changed
This PR adds support for Secp256r1/ES256. Reason is a need to sign a Proof of Possession JWT in the OpenID 4 Verifiable Credentials track of the W3C EDU Jobs for the Future Plugfest 2. Most OID4VCI issuer support ES256 and only some support ES256k. Since Veramo already support ES256k and has DID-JWT as a dep, which supports Secp256r1/ES256, this PR does not add any additional deps that weren't already transitive dependencies for KMS-Local anyway (elliptic package).

## Quality
Check all that apply:
* [X] I successfully ran `yarn`, `yarn build`, `yarn test`, `yarn test:browser` locally. (Yarn test fails on Ethereum tests, but that has nothinig todo with this PR)
* [X] I added unit tests.


mirceanis (changes requested):
Thank you for adding this!

There are a couple of things to resolve before this can be released successfully:

First one is the line endings being changed in some of the plugin-schema.json files. I think it will fail in the release process otherwise. This can also be fixed in another PR, if your machine doesn't allow it.

Second one is a clarification about `ES256-R`. Is it defined somewhere as a JWA? If it isn't we should remove it.

We've been criticized before for adding what looks like standards but are not, and I somewhat understand that position, so we shouldn't be creating extra confusion about algorithms.

nklomp:
Thanks for the quick feedback @mirceanis. You are correct that ES256-R is not standard, so removed it altogether.

Reverted the EOL changes in the plugin.schema.json. I did notice them, but hoped they wouldn't pose a problem. My IDE and the editor in Github does automatically add a newline at the end of the file though.

Another thing I noticed is that after running a yarn build-clean my plugin.json contained several 'description' lines that aren't present in the repo

mirceanis (approved):
looks great, thanks for adding this!

PR #1045, opened 2022-10-25
ION plugin fixes

nklomp:
- Intermittent test failures
- Make some errors more clear
- package.json not in line with the rest of Veramo (build dir, github etc)
- Refactored deleting rotation keys in case of errors

Fixed the intermittent ION tests failing. Has to do with Microsoft's ION node returning an error on resolution every now and then. This is an issue on MS side as the same DID works when refreshing. It also happens with DIDs anchored weeks ago.

Made the returned errors more clear, so we at least know what is going on.

The package.json had an output/build directory which isn't in line with Veramo. Also updated tags and the repo

Refactored deleting a rotation key when issues occur in the rest of the code. Move it to a method, that ignores deletion errors and debug logs them. Failure to remove the key isn't a serious problem for the rest of the code and we don't want to hide the original problem that caused the removal anyway.

## What is being changed
See above

## Quality
Check all that apply:
* [X ] I successfully ran `yarn`, `yarn build`, `yarn test`, `yarn test:browser` locally.
* [ ] I added unit tests.

## Details


mirceanis (approved):
Looks good. Thank you for the contribution!

PR #1053, opened 2022-11-03
feat: add did-provider-ens

nickreynolds:
## What is being changed
add a new DIDProvider to provide `did:ens` DIDs

## Quality
Check all that apply:
* [ ] I added unit tests.

## Details
related PR in https://github.com/veramolabs/ens-did-resolver/pull/12


nickreynolds:
I need to make sure to add testnet (goerli) support, before merging this

mirceanis:
> I need to make sure to add testnet (goerli) support, before merging this

If it's ok with you, I converted it to a draft PR until everything checks out.

Looks great so far!

mirceanis:
@nickreynolds what do you think is needed here to complete this PR?

nickreynolds:
@mirceanis I want to make sure we have support for `goerli` ens names. I can get back on this today/tomorrow

nickreynolds:
@mirceanis @simonas-notcat this is ready to merge. Was a draft PR before, so needs a review

packages/did-provider-ens/src/ens-did-provider.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  async createIdentifier(
+    { kms, alias, options }: { kms?: string; alias?: string; options: any },
+    context: IContext,
+  ): Promise<Omit<IIdentifier, 'provider'>> {
+    const keyType = options?.keyType || 'Secp256k1'
+    const key = await context.agent.keyManagerCreate({ kms: kms || this.defaultKms, type: keyType })
+
+    const identifier: Omit<IIdentifier, 'provider'> = {
```
mirceanis, line 28: In its current form, these identifiers wouldn't be usable for signing credentials, because the keys created here don't get reflected in the resolved DID document for the `did:ens`.
This pattern is currently being used by the `did-provider-web` package, but the assumption there is that the same agent managing the DID is the one that serves the DID document (using [@veramo/remote-server#WebDidDocRouter](https://github.com/uport-project/veramo/blob/fb22e632ef6dcce6a7dfec9a229c7be4d6d5c894/packages/remote-server/src/web-did-doc-router.ts#L40)). We should make this assumption more clear there too.

To get to that point we need to use [one of the ENS SDKs](https://docs.ens.domains/dapp-developer-guide/ens-libraries#javascript) to modify the actual text records mentioned in the [did:ens spec](https://github.com/veramolabs/did-ens-spec).

mirceanis (changes requested):
This is a good start, but in its current form it doesn't result in usable DIDs yet.

A halfway alternative solution is to only allow import for these kinds of DIDs along with their controller keys, putting the responsibility of maintaining the correct mapping on the user, and leaving the reservation of the ENS name to the existing ENS registrar UIs.

packages/did-provider-ens/src/ens-did-provider.ts
mirceanis: That's a good point, although this kind of race-condition is probably true of other DID methods related to ethereum as well.

PR #1064, opened 2022-11-10
fix: didcomm message handler should attempt to pass message to other handlers

### LINKED ISSUE #944: DIDComm message handler breaks out of message handler flow
Once a DIDComm v2 message is unpacked, no further messageHandlers are executed. Maybe this is intended behavior, but I was expecting to use subsequent message handlers to perform more processing on the message.

The fix would be easy, changing line 124 of the DidComm message-handler.ts from "return message" to "return super.handle(message, context)".

nickreynolds:

fixes #944

## What is being changed
DIDComm Message Handler should attempt to forward message to downstream handlers (but should not fail if none handle message)

## Quality
Check all that apply:
* [ ] I added unit tests.

### DIFF
--- a/__tests__/shared/didCommAndDataStoreWithCredentials.ts
+++ b/__tests__/shared/didCommAndDataStoreWithCredentials.ts
@@ -1,0 +1,193 @@
+// noinspection ES6PreferShortImport
+
+import {
+  IAgentOptions,
+  IDIDManager,
+  IEventListener,
+  IIdentifier,
+  IKeyManager,
+  IResolver,
+  TAgent,
+} from '../../packages/core/src'
+import { IDIDComm, IPackedDIDCommMessage } from '../../packages/did-comm/src'
+import { v4 } from "uuid"
+import { VerifiableCredential } from '@veramo/core'
+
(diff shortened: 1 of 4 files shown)


__tests__/shared/didCommAndDataStoreWithCredentials.ts
```diff
@@ -0,0 +1,193 @@
...
+      const verifiableCredential = await vc(creator, 'lds')
+      const packedMessage = await packed(verifiableCredential)
+
+      await agent.sendDIDCommMessage({        
+        messageId: 'test-jwt-success',
+        packedMessage,
+        recipientDidUrl: sender.did,
+      })
+      const messages = await agent.dataStoreORMGetMessagesCount({})
+      expect(messages).toEqual(2)
```
mirceanis, line 151: testing message flow by count is prone to some weird test failures when multiple tests produce messages.

It would be better to use message IDs, or something else to check if they get processed correctly.

mirceanis (changes requested):
This is a very nice fix that also doesn't introduce a breaking change, which I would have expected for #944. Great work!

The test cases can be improved a little bit; please see my comments.

PR #1074, opened 2022-11-23
Fix/ion tests await and update transmute did-key dep

### LINKED ISSUE #1066: didManagerCreate throws error for did:ion
**Bug severity**
4

**Describe the bug**
When attempting to create an ION DID with didManagerCreate, I get a "TypeError: crypto__WEBPACK_IMPORTED_MODULE_2___default(...).createHash is not a function" error.

Creating did:ethr and did:key work fine. My app uses the crypto module and createHash function elsewhere with no issues.

The error points to @transmute/did-key-secp256k1 -> publicKeyJwkFromPublicKeyHex -> getKid

The did-provider-ion package.json has the dependency fixed at "@transmute/did-key-secp256k1": "0.2.1-unstable.42", but 0.3.0 is current. Maybe that's the issue.

 - Veramo: next
 - Browser Brave
 - Node Version 16.17.0
 - Ionic 6.18.1
 - Capacitor 4.5.0
 - Angular 14.2.11

nklomp:
Intermittent timeout exceptions occur on ION test when running locally as reported by @nickreynolds  I had a hard time reproducing it using several environments, operating systems and node versions and actually managed to only get the error one time.

Looking at the test code it potently could happen because certain corner cases didn't await the matching methods which are async. Ran the tests many times over, and could not reproduce it anymore

Also removed the dep to transmute's did-key-jwk/secp256k1 support. The ION provider was using an older version of the dep than used elsewhere. The only reason was because it used 2 methods to convert secp256k private and public keys to JWKs. Since the structure of the package has been completely overhauled in recent versions, and these implementations changed, decided to copy the respective methods and use some previous transitive deps directly. 

Potential fixes #1066 . But I think that problem is not related to the mismatch in package versions

## What is being changed
See above

## Quality
Check all that apply:
* [x] I successfully ran `yarn`, `yarn build`, `yarn test`, `yarn test:browser` locally.
* [x] I added unit tests.


packages/did-provider-ion/src/functions.ts
```diff
@@ -48,9 +55,75 @@ export const toIonPrivateKeyJwk = (privateKeyHex: string): JwkEs256k => {
...
+/** convert compressed hex encoded public key to jwk */
+export const publicKeyJwkFromPublicKeyHex = (publicKeyHex: string) => {
+  let key = publicKeyHex;
+  const compressedHexEncodedPublicKeyLength = 66;
+  if (publicKeyHex.length === compressedHexEncodedPublicKeyLength) {
+    const keyBin = secp256k1.publicKeyConvert(
+      Buffer.from(publicKeyHex, 'hex'),
+      false
+    );
+    key = Buffer.from(keyBin).toString('hex');
```
mirceanis, line 112: We already have @ethersproject/* in the dependency tree (usually) so It would be better to use that for public key conversions instead of a new dependency.

Using this import `import { computePublicKey } from '@ethersproject/signing-key'`,
I think this would work to convert the key to uncompressed hex encoding:
```suggestion
    key = computePublicKey(`0x${publicKeyHex}`, false).substring(2)
```

This also removes the need for `Buffer` which requires special treatment on web.

packages/did-provider-ion/src/functions.ts
```diff
@@ -48,9 +55,75 @@ export const toIonPrivateKeyJwk = (privateKeyHex: string): JwkEs256k => {
...
+  const copy = { ...jwk } as any;
+  delete copy.d;
+  delete copy.kid;
+  delete copy.alg;
+  const digest = crypto
+    .createHash('sha256')
+    .update(canonicalize(copy)!, 'utf-8')
+    .digest();
+
+  return base64url.encode(Buffer.from(digest));
```
mirceanis, line 87:
```suggestion
  return u8a.toString(digest, 'base64url')
```

We prefer to avoid Buffer since it requires special treatment on web.
Perhaps with this change we also don't need `base64url` dependency anymore?

packages/did-provider-ion/src/functions.ts
```diff
@@ -48,9 +55,75 @@ export const toIonPrivateKeyJwk = (privateKeyHex: string): JwkEs256k => {
...
+ * See [rfc7638](https://tools.ietf.org/html/rfc7638) for more details on Jwk.
+ */
+export const getKid = (
+  jwk: ISecp256k1PrivateKeyJwk | ISecp256k1PublicKeyJwk
+) => {
+  const copy = { ...jwk } as any;
+  delete copy.d;
+  delete copy.kid;
+  delete copy.alg;
+  const digest = crypto
```
mirceanis, line 82: In other parts of the codebase we use `import { hash } from '@stablelib/sha256'` to compute sha256.

What do you think about replacing the use of `crypto`, since it requires special treatment on web?

mirceanis (changes requested):
Thanks for adding this!

I have some suggestions that would replace some of the dependencies with others that are already used in other parts of the codebase.
Please take a look and let me know what you think.

Also, my delayed review allowed a conflict to appear that has to be solved before we can merge. Sorry about that.

packages/did-provider-ion/src/functions.ts
nklomp: Removed Buffer

packages/did-provider-ion/src/functions.ts
nklomp: Updated it everywhere

nklomp:
Removed some packages and also removed usage of crypto and Buffer.
Updated to the latest ion-sdk. Should be browser compatible, but didn't get around to testing it yet (and probably won't this week)

mirceanis (approved):
Looks great, thank you for the prompt edits!

PR #1078, opened 2022-11-29
fix(utils): convert JWK with curv `Ed25519` to `X25519`

nickreynolds:
## What is being changed
allow `JsonWebKeys` with `crv = Ed25519` to be converted to `crv = X25519`

## Quality
Check all that apply:
* [X] I added unit tests.

### DIFF
--- a/packages/did-comm/src/__tests__/packing.test.ts
+++ b/packages/did-comm/src/__tests__/packing.test.ts
@@ -134,6 +134,63 @@ const hexDoc = {
   ]
 }
 
+// https://github.com/aviarytech/didcomm/blob/master/tests/fixtures/didDocs/alice.json
+const jwkDocX = {
+	"@context": ["https://www.w3.org/ns/did/v1", "https://w3id.org/security/suites/jws-2020/v1"],
+	"id": "did:example:alice",
+	"verificationMethod": [
+		{
+			"id": "did:example:alice#key-0",
+			"controller": "did:example:alice",
+			"type": "JsonWebKey2020",
+			"publicKeyJwk": {
+				"kty": "OKP",
+				"crv": "X25519",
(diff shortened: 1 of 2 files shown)


mirceanis (approved):
This looks great!
Do you plan to fill the `todo` test cases as well?

nickreynolds:
I would like to add tests for all the `unpack` cases as well, but at least for JWK, I wouldn't know how to do that. We don't currently support importing JWKs into KeyManager, do we?

mirceanis:
> We don't currently support importing JWKs into KeyManager, do we?

No, we don't. Private key data has to be converted to hex. We should be able to do the conversion with existing utils/deps:
For JWK ec private keys this means converting the `d` property from base64 to base16.

nklomp:
In our SSI Wallet we have code that actually has support to covert DID jwk and thus JWKs to hex. We intent to bring that to Veramo, but maybe you can already use it

 https://github.com/Sphereon-Opensource/ssi-mobile-wallet/blob/79b7cb56a3d009f741463fb0ca0fb7997c1a090f/src/services/identityService.ts#L132


Called from https://github.com/Sphereon-Opensource/ssi-mobile-wallet/blob/79b7cb56a3d009f741463fb0ca0fb7997c1a090f/src/services/identityService.ts#L165

The last 3 methods are basically delegating to Veramo and then have specific support for JWKs

nklomp:
Btw we also have a did jwk provider and resolver https://github.com/Sphereon-Opensource/ssi-sdk/tree/develop/packages/jwk-did-provider

nickreynolds:
> In our SSI Wallet we have code that actually has support to covert DID jwk and thus JWKs to hex. We intent to bring that to Veramo, but maybe you can already use it
Yes what you have seems to be essentially what this PR provides, with the addition of the `p256` curve support, correct? Would love to bring that into Veramo

The `did-jwk-provider` also seems like it would help a lot with unit tests

PR #1080, opened 2022-11-30
feat(did-comm): add trust ping protocol

### LINKED ISSUE #1079: [proposal] Add Trust Ping Protocol
Implement the functionality of the Trust Ping Protocol, as defined here: https://identity.foundation/didcomm-messaging/spec/#trust-ping-protocol-10

A helper function for sending Trust Pings (function should accept recipient DID, and handle packing and sending of correctly defined DIDCommMessage) as well as a message handler for processing Trust Pings (and sending Trust Ping Responses, when required, which is the standard flow)

Work in progress here: https://github.com/uport-project/veramo/tree/nickreynolds/trust-ping

**Additional context**
Implementation from Aviary Tech here: https://github.com/aviarytech/didcomm/blob/master/src/lib/protocols/trust-ping/2.0/ping.ts

nickreynolds:
closes #1079 

## What is being changed

## Quality
Check all that apply:
* [ ] I successfully ran `yarn`, `yarn build`, `yarn test`, `yarn test:browser` locally.
* [x] I added unit tests.

## Details

### DIFF
--- a/__tests__/shared/didCommPacking.ts
+++ b/__tests__/shared/didCommPacking.ts
@@ -68,7 +68,6 @@ export default (testContext: {
         packing: 'none',
         message,
       })
-      console.log("packedMessage: ", packedMessage)
       expect(JSON.parse(packedMessage.message).typ).toEqual('application/didcomm-plain+json')
       const unpackedMessage = await agent.unpackDIDCommMessage(packedMessage)
       expect(unpackedMessage).toEqual({
@@ -92,7 +91,6 @@ export default (testContext: {
         packing: 'jws',
         message,
       })
-      console.log("packedMessage: ", packedMessage)
       const unpackedMessage = await agent.unpackDIDCommMessage(packedMessage)
       expect(unpackedMessage).toEqual({
         message,
--- a/__tests__/shared/didCommPacking.ts
+++ b/__tests__/shared/didCommPacking.ts
@@ -57,7 +57,7 @@ export default (testContext: {
     afterAll(testContext.tearDown)
 
     it('should pack and unpack a plaintext message', async () => {
-      expect.assertions(1)
+      expect.assertions(2)
       const message = {
         type: 'test',
         to: receiver.did,
@@ -68,6 +68,8 @@ export default (testContext: {
         packing: 'none',
         message,
       })
+      console.log("packedMessage: ", packedMessage)
+      expect(JSON.parse(packedMessage.message).typ).toEqual('application/didcomm-plain+json')
       const unpackedMessage = await agent.unpackDIDCommMessage(packedMessage)
       expect(unpackedMessage).toEqual({
         message: {
@@ -90,6 +92,7 @@ export default (testContext: {
         packing: 'jws',
         message,
       })
+      console.log("packedMessage: ", packedMessage)
       const unpackedMessage = await agent.unpackDIDCommMessage(packedMessage)
       expect(unpackedMessage).toEqual({
         message,
--- a/packages/did-comm/src/__tests__/message-handler.test.ts
+++ b/packages/did-comm/src/__tests__/message-handler.test.ts
@@ -17,7 +17,6 @@ import { Resolver } from 'did-resolver'
 import { DIDCommHttpTransport } from "../transports/transports"
 import { IDIDComm } from "../types/IDIDComm"
 import { MessageHandler } from "../../../message-handler/src"
-import { createTrustPingMessage, TrustPingMessageHandler } from "../protocols/trust-ping-message-handler"
 import { FakeDidProvider, FakeDidResolver } from "../../../test-utils/src"
 import { MessagingRouter, RequestWithAgentRouter } from '../../../remote-server/src'
 import { Entities, IDataStore, migrations } from '../../../data-store/src'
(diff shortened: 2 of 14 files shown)


packages/did-comm/src/__tests__/trust-ping-message-handler.test.ts
```diff
@@ -0,0 +1,230 @@
...
+
+   
+  it('should handle trust ping message sent via didcomm', async () => {
+    const trustPingMessage = createTrustPingMessage(sender.did, recipient.did)
+    const packedMessage = await agent.packDIDCommMessage({ packing: 'authcrypt', message: trustPingMessage})
+    const result = await agent.sendDIDCommMessage({ messageId: trustPingMessage.id, packedMessage, recipientDidUrl: recipient.did })
+    console.log("result: ", result)
+
+    // this counts the 2 calls from fisrt test, 4 from previous (and then 4 from current test), this is not a good way to check this!
+    expect(DIDCommEventSniffer.onEvent).toHaveBeenCalledTimes(10)
```
nickreynolds, line 227: this is currently failing, as recipient does not properly unpack message

packages/did-comm/src/__tests__/trust-ping-message-handler.test.ts
```diff
@@ -0,0 +1,230 @@
...
+    //   { type: 'DIDCommV2Message-received' },
+    //   expect.anything(),
+    // )
+    
+    // // expect(handled.metaData).
+    // expect(DIDCommEventSniffer.onEvent).toHaveBeenCalledWith(
+    //   { type: 'DIDCommV2Message-sent' },
+    //   expect.anything(),
+    // )
+    expect(DIDCommEventSniffer.onEvent).toHaveBeenCalledTimes(2)
```
nickreynolds, line 201: need a better check for these tests

packages/message-handler/src/abstract-message-handler.ts
```diff
@@ -19,6 +19,7 @@ export abstract class AbstractMessageHandler {
     if (this.nextMessageHandler) {
       return this.nextMessageHandler.handle(message, context)
     }
+    console.warn("Can't handle message: ", message)
```
mirceanis, line 22: We should be using the `debug` module for this kind of logging, since that can be switched off or filtered.

packages/did-comm/src/protocols/trust-ping-message-handler.ts
```diff
@@ -0,0 +1,82 @@
...
+   * Handles a Trust Ping Message
+   * https://identity.foundation/didcomm-messaging/spec/#trust-ping-protocol-10
+   */
+  public async handle(message: Message, context: IContext): Promise<Message> {
+    if (message.type === TRUST_PING_MESSAGE_TYPE) {
+      debug('TrustPing Message Received')
+      try {
+        const { from, to, id } = message
+        if (!from) {
+          throw new Error("Trust Ping Message received without `from` set")
```
mirceanis, line 57: Let's use a prefix for errors, for consistency.

In this case, something like `invalid_argument: Trust Ping Message received without "from" set`

packages/did-comm/src/__tests__/message-handler.test.ts
```diff
@@ -0,0 +1,226 @@
...
+      },
+      expect.anything(),
+    )
+  }
+
+  const getRegularMessage = () => {
+    return { id: v4(), type: 'fake', to: recipient.did, from: sender.did, body: { hello: 'world' }}
+  }
+
+  it('should pack and unpack trust ping message with none packing', async () => {
```
mirceanis, line 205: I think the test name should refer to regular message here instead, no?

__tests__/shared/didCommPacking.ts
```diff
@@ -68,6 +68,8 @@ export default (testContext: {
         packing: 'none',
         message,
       })
+      console.log("packedMessage: ", packedMessage)
```
mirceanis, line 71: If these logs were used for debug, they should be removed to reduce the noise in tests

mirceanis (changes requested):
Looks good, kudos for the thorough testing!

It's a straight forward approach, using the message handler chain for this, and consistent with the existing codebase.
I'm not sure if this is a problem for this protocol, but this can be triggered by any message that looks like a trust ping, even if it doesn't come through didcomm. 

Another possible approach for protocol implementations would be to create protocol handlers as agent plugins that react based on event listeners.
Maybe for future protocols we can also explore this option but we need to figure out how to forward events for agents that use `@veramo/remote-client`

That being said, now with a new message handler supported, we should also update our default CLI config file.

nickreynolds:
> It's a straight forward approach, using the message handler chain for this, and consistent with the existing codebase. I'm not sure if this is a problem for this protocol, but this can be triggered by any message that looks like a trust ping, even if it doesn't come through didcomm.
> 
> Another possible approach for protocol implementations would be to create protocol handlers as agent plugins that react based on event listeners.

@mirceanis 

Yeah, it being able to be triggered by "non didcomm messages" was intentional. It at least makes testing a bit easier since it can be separated from didcomm's own message handler. I don't feel strongly though, but it seems fine to do it this way?

I was considering using the event listener system as well, but it seemed to work well with the message handler flow. If we decide that other "didcomm protocols" should use the event system, I can refactor this one at that time.

Made the small changes you requested

mirceanis (approved):
Looks great!

Yeah, we can refactor to event system later, if needed.

Please also consider adding this new handler to the default CLI config

PR #1095, opened 2022-12-31
feat(did-comm) Support DIDComm Mediation

### LINKED ISSUE #1051: [proposal] Add Support for DIDCommv2 Mediation
### Background

The [DIDComm v2 Specification](https://identity.foundation/didcomm-messaging/spec/v2.0/) was approved by DIF a few months ago. While Veramo already has support for much of this version of the spec, it currently lacks support for the "mediation" concept, which can be especially useful for bringing DIDComm to end-user agents, which may not be always online or expose http endpoints

This issue represents the work required to implement and demonstrate the [DIDComm Messaging Routing Protocol](https://identity.foundation/didcomm-messaging/spec/#routing-protocol-20) as well as the [Mediator Coordination Protocol](https://didcomm.org/mediator-coordination/2.0/).
 
### Requirements

* [ ] - create a Veramo plugin that allows Veramo agents to act as DIDComm Mediators
* Plugin should be a MessageHandler that attempts to handle messages after they've been processed by DIDCommMessageHandler
* Plugin should expose some functionality to allow agent to "request mediation", update DIDDocument with Mediator info if mediation granted, and save info (on Mediator-side) required to eventually forward messages to requester (the Mediator Coordination Protocol)
* For this implementation, we can assume that the recipient will eventually send an http request to the mediator to check for messages

* [ ] - update `didcomm` plugin to support mediation via `forward messages` [https://github.com/uport-project/veramo/blob/next/packages/did-comm/src/didcomm.ts#L536](https://github.com/uport-project/veramo/blob/next/packages/did-comm/src/didcomm.ts#L536)
* Must support any number of "forward message" wrappers required. Should demonstrate sending a DIDComm message that passes through at least 2 mediators.
* Likely requires changes to `packDIDCommMessage` and `sendDIDCommMessage`. 

* [ ] - Demonstrate functionality in a browser-based Veramo agent using `did:ethr` to specify mediator, set up relationship with Mediator, and poll mediator (via http request) to receive messages.
* This can be added in a branch of [https://github.com/veramolabs/agent-explorer/](https://github.com/veramolabs/agent-explorer/)

### Testing

Extensive testing will be needed to cover both the process of sending a message to a recipient via a mediator, as well as the process of setting up the mediation relationship. 

* Agent 1 (DID1) is able to successfully "request mediation" from Mediator Agent, and update their DIDDocument to indicate routing DID
* Agent 2 is able to pack a message for DID1, wrap that message in a `forward` message, and send it to the Mediator Agent. 
* Agent 1 is able to receive their held messages from Mediator Agent. These messages should be handled by Agent 1's DIDCommMessageHandler

As many failure cases as reasonable should also be added.

While not a strict limit, implementers should aim for a code-diff-coverage of ~75%

### Constraints

* **Minimal new dependencies**: Packages in this repository are often installed together so dependencies must be reused between them when possible. No new dependencies should be introduced if this can be avoided.

### Resources
[https://didcomm.org/book/v2/routing](https://didcomm.org/book/v2/routing)
[https://didcomm.org/mediator-coordination/2.0/](https://didcomm.org/mediator-coordination/2.0/)

codynhat:
Initial part of #1051. Adds support for forwarding messages to mediators. I am submitting a draft PR here to make sure I am on the right track.

Still to do is handling forwarded messages and setting up the relationship between a user and mediator.

closes #1051 

## What is being changed

## Quality
Check all that apply:
* [x] I want these changes to be integrated
* [x] I successfully ran `yarn`, `yarn build`, `yarn test`, `yarn test:browser` locally.
* [x] I added unit tests.


packages/did-comm/src/didcomm.ts
```diff
@@ -494,15 +496,60 @@ export class DIDComm implements IAgentPlugin {
...
+        {
+          media_type: DIDCommMessageMediaType.ENCRYPTED,
+          data: {
+            json: JSON.parse(packedMessageToForward.message),
+          },
+        },
+      ],
+    }
+
+    context.agent.emit('DIDCommV2Message-forwarded', {
```
nickreynolds, line 532: we want to emit this message at some point, but probably not here, since this is just the message wrapping. this event shouldn't get emitted until after the message is actually forwarded.

packages/did-comm/src/didcomm.ts
```diff
@@ -494,15 +496,60 @@ export class DIDComm implements IAgentPlugin {
...
+      messageId,
+      next: recipientDidUrl,
+      routingKey: routingKey,
+    })
+
+    // 2. Pack message for routingKey with anoncrypt
+    if (shouldUseSpecificKid) {
+      return this.packDIDCommMessageJWE(
+        { message: forwardMessage, packing: 'anoncrypt', options: { recipientKids: [routingKey] } },
```
nickreynolds, line 541: does the forward protocol assume anoncrypt?

codynhat: I don't see anywhere in the spec where this is mentioned explicitly. The example does not have a `from` field, meaning anoncrypt is used. https://identity.foundation/didcomm-messaging/spec/#messages

This seems to mean either could be used?

packages/did-comm/src/didcomm.ts
codynhat: What about if the event was `DIDCommV2Message-wrappedForward`? In the case of their being more than one mediator, multiple messages would be wrapped but only one would be forwarded. Should this emit one message per wrap? Or only one message for the single forward?

codynhat:
@nickreynolds This is ready for review. Here is a summary:

- Three DIDComm protocols (Pickup, Mediator Coordinator, Routing) were added under `protocols` in the `didcomm` package, similar to Trust Ping
- Routing Protocol
  - Handles a mediator receiving `forward` messages and saving one message for each `kid`
- Mediator Coordinator Protocol
  - Handles a mediator receiving `mediate-request` and currently grants mediation to anyone who requests it
  - Handles a receiver receiving `mediate-grant` and `mediate-deny` and updates the DID doc service endpoint accordingly
- Pickup Protocol
  - Handles a mediator receiving a `status-request` message, currently responding with the required `message_count`
  - Handles a mediator receiving a `delivery-request` and responding with a `delivery`
  - Handles a mediator receiving a `messages-received` message and deleting messages
  - Handles a recipient receiving a `delivery` request and handling all messages in a batch
- Also adds `return_route` functionality to the HTTP transport
- Also adds wrapping of messages in a `forward` when sent to a DID with a mediator in their DID doc

nickreynolds:
At first glance, this looks amazing. Thank you so much @codynhat ! 

This one might take a little while to fully review (and we're trying to finish up the ESM upgrade before merging in any other big features so might go unmerged even after it's approved) but thanks!

codynhat:
Sounds good!

packages/data-store-json/src/data-store-json.ts
```diff
@@ -137,6 +138,18 @@ export class DataStoreJson implements IAgentPlugin {
     }
   }
 
+  async dataStoreDeleteMessage(args: IDataStoreDeleteMessageArgs): Promise<boolean> {
```
nickreynolds, line 141: @mirceanis I see that you originally removed this function, can you review this file's changes?

mirceanis: There were some unpleasant consequences when deleting stuff because of some typeorm misconfiguration in our entities, causing a cascade effect and deleting more than just the message, or throwing foreign key errors.

This should not be the case now, but this is the reason why some "delete" methods are weirdly missing or commented out.

The missing methods were not added to the data-store-json plugin to keep the API uniform, but we need to review this and normalize the API.

nickreynolds (approved):
Looks great!

PR #1128, opened 2023-02-20
add did:jwk method support

tadejpodrekar:
## What is being changed
Added an implementation of 'AbstractIdentifierProvider' for the did:jwk method according to the [specification](https://github.com/quartzjer/did-jwk/blob/main/spec.md). 
The provider currently supports Secp256k1, Secp256r1, Ed25519, X25519 key types.

* Add create did:jwk identifier tests
Passing

* Add resolve did:jwk tests
Passing

* Add tests creating a valid VC using the did:jwk method
Passing

## Quality
Check all that apply:
* [ ] I added unit tests.


packages/did-provider-jwk/src/jwkDidUtils.ts
```diff
@@ -0,0 +1,53 @@
...
+    return {
+      crv: 'secp256k1',
+      kty: 'EC',
+      x: bytesToBase64url(hexToBytes(x.toString('hex'))),
+      y: bytesToBase64url(hexToBytes(y.toString('hex'))),
+    } as JsonWebKey
+  }
+  if(keyType === 'Secp256r1') {
+    const EC = new elliptic.ec('p256')
+    // add '03' prefix to public key
```
mirceanis, line 22: Where is this prefix coming from?
Can you please share a link to some reference?

packages/did-provider-jwk/src/resolver.ts
```diff
@@ -0,0 +1,108 @@
...
+            type: 'JsonWebKey2020',
+            controller: did,
+            publicKeyJwk: jwk,
+          },
+        ],
+        assertionMethod: [`${did}#0`],
+        authentication: [`${did}#0`],
+        capabilityInvocation: [`${did}#0`],
+        capabilityDelegation: [`${did}#0`],
+        keyAgreement: [`${did}#0`],
```
mirceanis, line 48: According to the did:jwk spec, the key relationships should consider the `use` property of a JWK.
* When `use` is `enc`, only the `keyAgreement` relationship should be added.
* When `use` is `sig`, all except the `keyAgreement` relationship should be added.

I'm not aware of other implementations of this resolver so I can't tell if imposing this alignment with the spec would cause incompatibilities. What do you think?

packages/did-provider-jwk/src/resolver.ts
```diff
@@ -0,0 +1,108 @@
...
+      reject(error)
+    }
+  })
+}
+
+function parseDidJwkIdentifier(didIdentifier: string): JsonWebKey {
+  try {
+    const jwk = JSON.parse(decodeBase64url(didIdentifier)) as unknown
+    if (!isJWK(jwk)) {
+      throw new Error("DID identifier doesn't contain a valid JWK")
```
mirceanis, line 61: We try to prefix error messages for easier interpretation.
In the case of this resolver, all errors fall under an `illegal_argument: ` prefix

```suggestion
      throw new Error("illegal_argument: DID identifier doesn't contain a valid JWK")
```

packages/did-provider-jwk/src/resolver.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+    return jwk
+  } catch (error) {
+    throw new Error('Invalid DID identifier')
```
mirceanis, line 65:
```suggestion
    throw new Error('illegal_argument: Invalid DID identifier')

```

packages/did-provider-jwk/src/resolver.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+export const resolveDidJwk: DIDResolver = async (
+  did: string,
+  parsed: ParsedDID,
+  resolver: Resolvable,
+  options: DIDResolutionOptions,
+): Promise<DIDResolutionResult> => {
+  try {
+    if (parsed.method !== 'jwk') throw Error('Invalid DID method')
```
mirceanis, line 76:
```suggestion
    if (parsed.method !== 'jwk') throw Error('illegal_argument: Invalid DID method')

```

packages/did-provider-jwk/src/resolver.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    const didIdentifier = did.split('did:jwk:')[1]
+    if (!didIdentifier) throw Error('Invalid DID')
```
mirceanis, line 79:
```suggestion
    if (!didIdentifier) throw Error('illegal_argument: Invalid DID')

```

packages/did-provider-jwk/src/jwk-did-provider.ts
```diff
@@ -0,0 +1,107 @@
+import { IIdentifier, IKey, IService, IAgentContext, IKeyManager } from '@veramo/core-types'
+import { AbstractIdentifierProvider } from '@veramo/did-manager'
+import { VerificationMethod } from 'did-resolver'
+import type { JwkDidSupportedKeyTypes } from '.'
```
mirceanis, line 4: Perhaps it's easier to declare the type here, or in its own file instead of in the index file. There's less potential for circular dependency issues that way.

packages/did-provider-jwk/src/jwkDidUtils.ts
```diff
@@ -0,0 +1,53 @@
...
+    } as JsonWebKey
+  }
+  if(keyType === 'Ed25519') {
+    return {
+      crv: 'Ed25519',
+      kty: 'OKP',
+      x: bytesToBase64url(typeof pubKey === 'string' ? hexToBytes(pubKey) : pubKey),
+    } as JsonWebKey
+  }
+  throw new Error('Failed to create JWK')
```
mirceanis, line 48:
```suggestion
  throw new Error(`not_supported: Failed to create JWK using ${keyType}`)
```

packages/did-provider-jwk/src/jwk-did-provider.ts
```diff
@@ -0,0 +1,107 @@
...
+  async updateIdentifier(
+    args: {
+      did: string
+      kms?: string
+      alias?: string
+      options?: any
+    },
+    context: IAgentContext<IKeyManager>
+  ): Promise<IIdentifier> {
+    throw new Error('JwkDIDProvider updateIdentifier not supported')
```
mirceanis, line 61: We try to prefix error messages for easier interpretation.

All the errors in this file refer to operations that are not possible. We used `not_supported: ` as a prefix for this in other places.

```suggestion
    throw new Error('not_supported: JwkDIDProvider updateIdentifier not possible')
```

packages/did-provider-jwk/src/jwk-did-provider.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  async addKey({
+      identifier,
+      key,
+      options,
+    }: { identifier: IIdentifier; key: IKey; options?: any },
+    context: IContext
+  ): Promise<any> {
+    throw Error('JwkDIDProvider addKey not supported')
```
mirceanis, line 81:
```suggestion
    throw Error('not_supported: JwkDIDProvider addKey not possible')
```

packages/did-provider-jwk/src/jwk-did-provider.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  async addService({
+      identifier,
+      service,
+      options,
+    }: { identifier: IIdentifier; service: IService; options?: any },
+    context: IContext
+  ): Promise<any> {
+    throw Error('JwkDIDProvider addService not supported')
```
mirceanis, line 91:
```suggestion
    throw Error('not_supported: JwkDIDProvider addService not possible')
```

packages/did-provider-jwk/src/jwk-did-provider.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  async removeKey(
+    args: { identifier: IIdentifier; kid: string; options?: any },
+    context: IContext
+  ): Promise<any> {
+    throw Error('JwkDIDProvider removeKey not supported')
```
mirceanis, line 98:
```suggestion
    throw Error('not_supported: JwkDIDProvider removeKey not possible')

```

packages/did-provider-jwk/src/jwk-did-provider.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  async removeService(
+    args: { identifier: IIdentifier; id: string; options?: any },
+    context: IContext
+  ): Promise<any> {
+    throw Error('JwkDIDProvider removeService not supported')
```
mirceanis, line 105:
```suggestion
    throw Error('not_supported: JwkDIDProvider removeService not possible')

```

mirceanis (changes requested):
An excellent PR, thank you for contributing this!

I added some notes about error messages and a question about the spec.
Please take a look.

packages/did-provider-jwk/src/jwkDidUtils.ts
nklomp: We do it in our implementation as well (compressed format):
https://github.com/Sphereon-Opensource/ssi-sdk/blob/develop/packages/jwk-did-provider/src/functions.ts#L113

nklomp:
Cool. We also created a JWK DID provider for Veramo, which looks very similar but has some small changes compared to this PR.

- Also allows to import keys, instead of relying on key generation
- Takes the use property of a JWK into account
- resolution is more compliant to the general DID spec with errors and resolution metadata

We created it in our repo, as we needed it as part of the JFF Plugfest. I think it makes sense that we merge some features.

packages/did-provider-jwk/src/jwkDidUtils.ts
tadejpodrekar: It is added to indicate a compressed public key, otherwise we have an invalid hex string. The prefix (02 if the `y` coordinate of the public key point is even, 03 if the `y` is odd) is removed in the Veramo kms: https://github.com/uport-project/veramo/blob/abfba31ff8497f00cddec5ca7a561fe32b1ab0bb/packages/kms-local/src/key-management-system.ts#L333

packages/did-provider-jwk/src/jwk-did-provider.ts
tadejpodrekar: I have moved the type declarations to a separate file.

packages/did-provider-jwk/src/resolver.ts
tadejpodrekar: I have added the `use` property to the resolved JWK object and the corresponding key relationships according to the spec.

tadejpodrekar:
@mirceanis thank you for your comments. I have updated the implementation according to your suggestions. The changes also contain support for better error handling for the did resolution and the ability to import supported keys. The implementation is a bit different than the one from Sphereon, but the functionalities are covered.

@nklomp thank you very much for pointing that out. This PR should provide the basic functionalities for did:jwk method, improvements and changes to the implementation could be made in subsequent PR. If you have any other suggestions we are open to it.

packages/did-provider-jwk/src/jwkDidUtils.ts
mirceanis: hmm.. this seems odd (no pun intended :) )
Veramo kms should not remove this prefix. At most, it should remove the `0x` prefix from hex encoded public keys, but not the parity.
If it's removing the parity it is a bug.

mirceanis (approved):
This looks good!

I still have doubts about the prefix added to Secp256r1 public keys and I'll have to look into the key manager implementation to figure out where it's removing the parity.

Did you already spot this to speed up my search?

packages/did-provider-jwk/src/jwkDidUtils.ts
mirceanis: Oh, it really is removing the parity (https://github.com/uport-project/veramo/blob/9c73d98fd217ed9a612767f49a235cdbf43619cb/packages/kms-local/src/key-management-system.ts#L333).

@nklomp I missed this in my original review. Can you share some insight for why this would be needed?

nklomp: It shouldn't. I think we might need to look at how to fix it for any keys already stored though

mirceanis: I can add a new script to the `@veramo/data-store#migrations`, but it will only cover the TypeORM based storage.

mirceanis: oh.. actually, it will have to be done manually because the TypeORM migrations don't have access to the decryption key to get the private keys and recompute the correct public key :(

nklomp: Yeah, I think we can do it in code as well. A simple check for the prefix on a Secp256r1 key should suffice

mirceanis: I created a PR with the fix and proposed a method to recompute the keys: #1137

I'd appreciate your review @nklomp

Issue #1139, opened 2023-02-27
Incorrect Ed25519VerificationKey publicKeyMultibase conversion


DaevMithran:
**Bug severity: 3**

**Describe the bug**
Conversion of publicKeyMultibase to publicKeyHex in Ed25519VerificationKey2020 doesn't remove the Multibase header present in the publicKeyMultibase according to the spec

**To Reproduce**
Steps to reproduce the behaviour:
1. Issue a JsonLD credential using an assertion method of type Ed25519VerificationKey2020 in a Issuer DID. This will result in an error. 

**Observed behaviour**
This is the below observed behviour, as the prefix is not removed the publicKeyHex is different from the local wallet keys
```  
throw Error(`key_not_found: No suitable signing key found for ${identifier.did}`);
                  ^
Error: key_not_found: No suitable signing key found for did:cheqd:testnet:570996eb-393a-4659-8f88-71cb53488eb4
    at CredentialIssuerLD.findSigningKeyWithId 
```
   
**Expected behaviour**
The credential should be issued to the holder

2. Verify JWT credential issued by Ed25519VerificationKey2020 method Issuer DID.

**Observed behaviour**
 The credential was not verified successfully
 
 **Expected behaviour**
 The credential was verified successfully

PR #1082, opened 2022-12-02
feat(remote-server): use 2020 key types in web did doc router

nickreynolds:
## What issue is this PR fixing
`did:web` DID Documents hosted on Veramo web servers previously displayed Ed25519 Verification Keys in 2018 format, with outdated `publicKeyHex`. Moving to the 2020 version, with `publicKeyMultibase` should provide more interoperability

## What is being changed
* Add test to exercise Ed25519Signature2020 Suite
* Build DID Doc in `web-did-doc-router` with 2020 keys
* moved a helper function from `credential-ld/src/suites/Ed25519Signature2020.ts` to `utils/src/encodings.ts`
 
## Quality
Check all that apply:
* [X] I want these changes to be integrated
* [ ] I successfully ran `yarn`, `yarn build`, `yarn test`, `yarn test:browser` locally.
* [ ] I allow my PR to be updated by the reviewers (to speed up the review process).
* [X] I added unit tests.
* [ ] I added integration tests.
* [ ] I did not add automated tests because _________, and I am aware that a PR without tests will likely get rejected.

## Details
If applicable, add screen captures, error messages or stack traces to help explain your problem.

PR #1146, opened 2023-03-14
fix(key): Isolate byte conversion fix to utils scope

Eengineer1:
FYI, #1139 .


packages/utils/src/did-utils.ts
```diff
@@ -138,7 +138,7 @@ function extractPublicKeyBytes(pk: VerificationMethod): Uint8Array {
   if (pk.publicKeyBase58) {
     return base58ToBytes(pk.publicKeyBase58)
   } else if (pk.publicKeyMultibase) {
-    return bases['base58btc'].decode(pk.publicKeyMultibase)
```
nickreynolds, line 141: this line should be refactored out to a utility function, similar to `hexToMultibase`. Should probably live in the same file as that function, `packages/utils/src/encodings.ts`

Eengineer1:
@mirceanis @nickreynolds 
Please, do take a look.
Fixed key conversion issues where relevant in tests.

Should unblock #1082 as well.

Apologies if belated, in comparison to complexity of this.

mirceanis:
Hi @Eengineer1, I'm still trying to find a proper spec for `publicKeyMultibase`.

It's unclear to me if it's supposed to be a multibase encoding of a multicodec encoding of a public key.

The encoding/decoding we were doing has likely been ported over from did:key implementations for Ed25519 keys, where multicodec is indeed needed in the identifier to be able to know what kind of key it is.

But verification methods already describe the key type, so there should be no need for multicodec.

This also means that if we keep using this `0xed` multicodec header for generic keys we are going to be corrupting keys of other types.
 
I'll keep digging. If you (or anyone else reading this) are also looking into this, please post links here.

packages/utils/src/encodings.ts
```diff
@@ -143,3 +144,52 @@ export function base58ToBytes(s: string): Uint8Array {
...
+/**
+ * Converts a multibase string to the Uint8Array it represents.
+ * 
+ * @param s - the string to be converted
+ * 
+ * @throws if the string is not formatted correctly.
+ * 
+ * @public
+*/
+export function multibaseToBytes(s: string): Uint8Array {
```
nickreynolds, line 157: https://www.w3.org/community/reports/credentials/CG-FINAL-di-eddsa-2020-20220724/#ed25519verificationkey2020

So according to the Ed25519VerificationKey2020 definition, this is correct. But this function assumes that it's a Ed25519VerificationKey2020  publicKeyMultibase, when it should probably take in the key type as a parameter so it can use different multicodec prefix (if needed) for other key types.

As far as I can tell, X25519KeyAgreementKey2020 has no definition, but we should assume it uses the `x25519-pub` multicodec prefix (0xec). I'd probably be ok saying the X25519 key can use either the x25519-pub OR ed25519-pub prefixes (at least in multibaseToBytes), since I've seen a few X25519 keys that were incorrectly constructed that way.

But bytesToMultibase also needs to be refactored to take in a keyType so we know which multicodec prefix to prepend.

PR #1152, opened 2023-03-28
Fix react-native migrations on older Android installations

nklomp:

In React-Native on older Android APIs (<11) the migrations where failing. That has to do with the 2nd and 3rd migration file. The 2nd migration file wasn't correctly creating new temp tables including privateKeyHex, because it picked up the wrong entity it seems. Probably because the matching logic on the metadata against a list of metadata has a different order in these environments.  As a result the 3rd migration also failed. It is now using a name when creating new tables, but using the actual matched table from the matching logic on existing tables.

Unrelated, the TypeORM migrations are failing in production settings where minifying is being used. That has to do with the fact that TypeORM relies on classnames and its timestamp suffix. When minifying this goes away, so I simply added the name property to the migration files.

As the same method was being copied over between migrations, also create 2 helper functions, so that the code is now shared across migrations and can also be used by people extending Veramo, which might have their own migrations.

Still needs more testing before converting this from draft to final PR

## What is being changed

## Quality
Check all that apply:
* [X] I added unit tests.

### DIFF
--- a/packages/data-store/src/migrations/migration-functions.ts
+++ b/packages/data-store/src/migrations/migration-functions.ts
@@ -1,0 +1,48 @@
+import { QueryRunner, Table } from 'typeorm'
+
+/**
+ * Get an existing table by name. Checks against givenTableName first, and tableName next. Throws an error if not found
+ *
+ * @param queryRunner The query runner object to use for querying
+ * @param givenName The given name of the table to search for
+ *
+ * @public
+ */
+export function migrationGetExistingTableByName(queryRunner: QueryRunner, givenName: string): Table {
+  const table = migrationGetTableByNameImpl(queryRunner, givenName)
+  if (!table) {
+    throw Error(`Could not find table with name ${givenName}`)
+  }
+  return table
+}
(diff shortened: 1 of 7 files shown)


packages/data-store/src/migrations/migration-functions.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Get an existing table by name. Checks against givenTableName first, and tableName next. Throws an error if not found
+ *
+ * @param queryRunner The query runner object to use for querying
+ * @param givenName The given name of the table to search for
+ *
+ * @public
+ */
+export function migrationGetExistingTableByName(queryRunner: QueryRunner, givenName: string, strictClassName?: boolean): Table {
```
mirceanis, line 11: Thanks for adding this. It was in the back of my head for a long time :)

mirceanis (approved):
Fantastic, thanks for adding this!

mirceanis (approved):
It looks great. Thank you for contributing this!

PR #1154, opened 2023-04-04
some Ed25519Signature2020 fixes (still broken)

mirceanis:
This PR adds some code to handle some of the quirks of the Ed25519Signature2020 suite.

* There was a bug in the signature encoding. It was being interpreted as utf-8 instead of base64url !!
* The @context gets fixed before signing/verifying
* The VerificationMethod public key material is re-encoded as publicKeyMultibase
* The documentLoader was updated to support resolving DID document fragments as well. Since the document loader is only setup once for the whole plugin, this will end up failing all the other signature suite tests :(

This PR doesn't fix the suite, but perhaps someone can take over for the following days while I'm OOO.

### DIFF
--- a/packages/credential-ld/src/ld-credential-module.ts
+++ b/packages/credential-ld/src/ld-credential-module.ts
@@ -10,6 +10,7 @@ import {
 import fetch from 'cross-fetch'
 import Debug from 'debug'
 import jsonldSignatures from '@digitalcredentials/jsonld-signatures'
+
 const { extendContextLoader } = jsonldSignatures
 import * as vc from '@digitalcredentials/vc'
 import { LdContextLoader } from './ld-context-loader.js'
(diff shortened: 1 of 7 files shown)


packages/credential-ld/src/ld-credential-module.ts
```diff
@@ -35,7 +36,11 @@ export class LdCredentialModule {
     this.ldSuiteLoader = options.ldSuiteLoader
   }
 
-  getDocumentLoader(context: IAgentContext<IResolver>, attemptToFetchContexts: boolean = false) {
+  getDocumentLoader(
+    context: IAgentContext<IResolver>,
+    attemptToFetchContexts: boolean = false,
+    resolveFragments: boolean = false,
```
mirceanis, line 42: setting this to true will unblock one of the errors in the 2020 suite, but will cause issues in all the others

packages/credential-ld/src/suites/Ed25519Signature2020.ts
```diff
@@ -38,8 +39,9 @@ export class VeramoEd25519Signature2020 extends VeramoLdSignature {
           keyRef: key.kid,
           data: messageString,
           encoding: 'base64',
+          algorithm: 'EdDSA',
         })
-        return u8a.fromString(signature)
+        return base64ToBytes(signature)
```
mirceanis, line 44: This kind of bug should not make it into our codebase..
We'll be much stricter about tests for new code going forward

PR #1176, opened 2023-05-11
fix: remove JSON.stringify for string credential

martines3000:
Fixes issue where a credential in string format is encoded as a string twice. 

## What is being changed
Removes JSON.stringify for string credential.

## Quality
Check all that apply:
* [ ] I added unit tests.

### DIFF
--- a/packages/credential-eip712/src/agent/CredentialEIP712.ts
+++ b/packages/credential-eip712/src/agent/CredentialEIP712.ts
@@ -218,7 +218,7 @@ export class CredentialIssuerEIP712 implements IAgentPlugin {
         if (typeof cred !== 'string' && cred.proof.jwt) {
           return cred.proof.jwt
         } else {
-          return JSON.stringify(cred)
+          return cred
         }
       })
       presentation.verifiableCredential = credentials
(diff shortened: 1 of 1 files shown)


packages/credential-eip712/src/agent/CredentialEIP712.ts
```diff
@@ -218,7 +218,7 @@ export class CredentialIssuerEIP712 implements IAgentPlugin {
         if (typeof cred !== 'string' && cred.proof.jwt) {
           return cred.proof.jwt
         } else {
-          return JSON.stringify(cred)
+          return cred
         }
```
mirceanis, line 222: There are 3 cases here:
* credential is already a string (JWT). return it directly
* credential is an Object but it's a JwtProof2020 proof, which is only used for convenience. The actual credential is in `cred.proof.jwt`, so return that instead
* credential is an Object with any other JSON-LD/EIP712 proof. In this case return `JSON.stringify(credential)`

The reason for this extra processing is that EIP712 proofs don't support heterogenous objects in an array, so the "quick fix" is to turn them all into strings.

PR #1180, opened 2023-05-18
feat(did-comm): add support for some of the AES based content and key encryption algorithms

### LINKED ISSUE #1083: [proposal] Add AES and NIST curve support to the DIDComm implementation
We're currently using only X25519 keys with ECDH-ED or 1PU + XC20P for key agreement and wrapping, and XC20P for content encryption, but XC20P is only listed as optional:
https://identity.foundation/didcomm-messaging/spec/#curves-and-content-encryption-algorithms

We need AES support for content encryption and key wrapping and P256 & P384 key support too. to be fully spec compliant in terms of algorithms.

It's likely that these features will have to be implemented in the lower level did-jwt library which is being used for JWE.
I believe that `@stablelib` (the lib that is used for encryption support) should already have all the primitives needed.

mirceanis:

partially fixes #1083

## What is being changed

Most relevant changes are in the did-comm package, where the new capabilities of the `did-jwt@7.2.0` are used to create a wider set of encrypters and logic to pick the proper decrypter too.
This PR adds support for `A256KW` key wrapping and for `A256GCM` and `A256CBC-HS512` content encryption.

These can be set as options to `packDIDCommMessage`. Example:
```ts
const packed = agent.packDIDCommMessage({
  packing: 'authcrypt',
  message,
  enc: 'A256GCM', // default for `authcrypt` or `anoncrypt`
  alg: 'ECDH-1PU+A256KW' // default for `authcrypt`
}),
```

There are some corner cases that may require an API review for the packing method:
`authcrypt` implies `ECDH-1PU`, so specifying an `ECDH-ES` alg will get overridden.
`anoncrypt` implies `ECDH-ES` so specifying an `ECDH-1PU` alg will be overridden.

The only supported key agreement type for now is `X25519`. Besides supporting other key types, there is one more missing piece to full did-comm v2 compliance:
* our `ECDH-1PU` implementation is still using the v3 draft of the spec, while the did-comm spec points to the v4 draft. The [diff between the 2 versions](https://author-tools.ietf.org/iddiff?url1=draft-madden-jose-ecdh-1pu-03&url2=draft-madden-jose-ecdh-1pu-04&difftype=--hwdiff) suggests they will be incompatible because of the `cctag` (section 2.3)

## Quality
Check all that apply:
* [X] I added unit tests.

## Details
This PR brings many seemingly unrelated upgrades but they are necessary due to the complicated dependency tree:
* `@stablelib/*` dependencies were replaced with `@noble/*` where available (similar to `did-jwt`)


packages/did-provider-jwk/src/jwkDidUtils.ts
```diff
@@ -54,33 +55,28 @@ function createJWK(
...
+        const point = secp256k1.ProjectivePoint.fromHex(pubKey).toAffine()
 
         return {
           alg: 'ES256K',
           crv: 'secp256k1',
           kty: 'EC',
           ...(keyUse && { use: keyUse }),
-          x: bytesToBase64url(hexToBytes(x.toString('hex'))),
-          y: bytesToBase64url(hexToBytes(y.toString('hex'))),
+          // FIXME: test endianness of the toString(16) output
```
mirceanis, line 65: Need to check this before merging this PR

mirceanis: It works fine. A test was added for this

__tests__/initial.migration.test.ts
```diff
@@ -167,7 +167,7 @@ describe('database initial migration tests', () => {
           encoding: 'utf-8',
         })
         expect(signedMessage).toEqual(
-          'vzDocUViJh7ooOCZ-jBHKZddEsTa4yClHwhIL9SHJwjAv3bC6TZIcUnX36ZqNBWvLbnNAQvdtzqrVf3l0pv3QQ',
```
mirceanis, line 170: JWT signing for ES256K now uses canonical signatures, with a low `S` value (through the use of `@noble/curves`)
This eliminates signature malleability, just like what happens for ethereum signatures, but it also means tests will change.

packages/data-store/src/__tests__/entities.test.ts
```diff
@@ -2,11 +2,9 @@ import { createCredentialEntity, Credential } from '../entities/credential.js'
 import { createPresentationEntity } from '../entities/presentation.js'
 import { Claim, Entities, Identifier, Message } from '../index.js'
 import { DataSource, In } from 'typeorm'
-// @ts-ignore
-import pkg from 'blakejs'
 import * as fs from 'fs'
-
-const { blake2bHex } = pkg
+import { bytesToHex } from '../../../utils/src'
```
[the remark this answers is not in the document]
mirceanis: Yes. I noticed a while ago that relative imports in test files work better while iterating on the file being imported.
If I were to import using `@veramo/utils` I'd have to rebuild before each test.

`jest` technically has a better way of automatically remapping these imports (using the `moduleMapper` config) but that did not work for me.

PR #1202, opened 2023-07-20
Use `keyRef` if supplied to pick signing key in `createVerifiablePres…

jasny:

https://github.com/uport-project/veramo/blob/6981e6845d6f90d7b0c8dd6e0117b73dfd8edfb9/packages/credential-w3c/src/action-handler.ts#L240-L242

`keyRef` is an arg property for `createVerifiablePresentation` and `createVerifiableCredential`. It should be used to pick which key is used for signing. However, it's being ignored.

## What is being changed
If `keyRef` is supplied, it's used to select a specific key of the identifier. If it's not supplied, the uses the first key (current behavior).

## Quality
Check all that apply:
* [x] I added unit tests.


mirceanis (approved):
looks good, kudos for updating the tests!

PR #1207, opened 2023-07-28
Allow DID parameters for verifiable credentials and presentations

### LINKED ISSUE #1201: [proposal] Support `versionTime` parameter when issuing a verifiable credential
When verifying a credential, the issuer id is resolved as-is. When the id is a DID `did:example:foo`, the credential is verified against the current version of the DID document. If a verification method is removed, all credentials signed with that key will become invalid. Depending on the use case, this can be the intended behavior or can be problematic.

The DID standard defines the [`versionTime` parameter](https://www.w3.org/TR/did-core/#did-parameters) which resolves to the version of the DID document that was valid at a certain time. Typically, using `versionTime` the result is immutable.

While not defined specifically like this in the VC standard, the `versionTime` parameter can be added to the `issuer` id to ensure that VCs aren't implicitly revoked when the DID document is modified. This should work with all verifiers, as they'll resolve the identifier as-is and should remove any parameters.

```js
{
  credential: {
    issuer: { id: `${identifier.did}?versionTime=${versionTime}` },
    credentialSubject: {
      id: 'did:web:example.com',
      you: 'Rock',
    },
  },
  proofFormat: 'jwt',
}
```

----

In `createVerifiableCredential`, the issuer is extracted from the verifiable credential and used to look up the identifier from the manager.

https://github.com/uport-project/veramo/blob/6981e6845d6f90d7b0c8dd6e0117b73dfd8edfb9/packages/credential-w3c/src/action-handler.ts#L209-L220

It should remove the DID parameters (or just `versionTime`) when looking up the identifier.

----

The alternative would be to always add `versionTime` when resolving a DID for verifying a credential. However, the standard doesn't indicate that that's the correct behavior.

jasny:
Fixes #1201

```
{
  credential: {
    issuer: { id: `${identifier.did}?versionTime=${versionTime}` },
    credentialSubject: {
      id: 'did:web:example.com',
      you: 'Rock',
    },
  },
  proofFormat: 'jwt',
}
```

Remove DID parameters in the `issuer` and `holder` properties when looking up the identifier using the DID manager. The DID manager doesn't handle DID parameters like `versionTime`. When the DID manager is used we should be able to safely ignore them. They're only used when resolving the DID.

## What is being changed

Added `options` property to `extractIssuer`, with option `removeParameters`. Added a `removeDIDParameters` function to credential-utils.

The `credentials-w3c`, `credentials-ld`, and `credentials-jwt` plugins all remove the DID parameters before looking up the identifier via the DID manager. If the DID is resolved, the parameters are not removed.

Modified `shortDid` to always remove parameters.

## Quality
Check all that apply:
* [X] I added unit tests.

No tests for:
* `credentials-jwt`, because there are no existing tests
* `credentials-ld` doesn't mock services, making it hard to test this feature

### DIFF
--- a/packages/cli/src/explore/utils.ts
+++ b/packages/cli/src/explore/utils.ts
@@ -3,6 +3,7 @@ import os from 'os'
 
 export function shortDid(did?: string): string {
   if (!did) return ''
+  did = did.replace(/\?.*$/, '')
   if (did.slice(0, 7) === 'did:web') {
     return did
   }
--- a/packages/credential-eip712/src/agent/CredentialEIP712.ts
+++ b/packages/credential-eip712/src/agent/CredentialEIP712.ts
@@ -14,6 +14,7 @@ import {
   MANDATORY_CREDENTIAL_CONTEXT,
   mapIdentifierKeysToDoc,
   processEntryToArray,
+  removeDIDParameters,
   resolveDidOrThrow,
 } from '@veramo/utils'
 import schema from "../plugin.schema.json" assert { type: 'json' }
@@ -63,7 +64,7 @@ export class CredentialIssuerEIP712 implements IAgentPlugin {
       issuanceDate = issuanceDate.toISOString()
     }
 
-    const issuer = extractIssuer(args.credential)
+    const issuer = extractIssuer(args.credential, { removeParameters: true })
     if (!issuer || typeof issuer === 'undefined') {
       throw new Error('invalid_argument: credential.issuer must not be empty')
     }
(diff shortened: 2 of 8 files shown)


mirceanis (approved):
Generally this looks good and everything seems to work just fine.
I'm a bit concerned that using a regex to strip away the query parameters might lead to some weird corner cases. I'd be more comfortable with using a URL Parser.

mirceanis:
We'll merge this as-is since the main use of the url-stripping is for matching against lists of other DIDs.

Thank you for the contribution @jasny

PR #1242, opened 2023-09-14
chore: bump ethers

fermentfan:
Still have to fix tests...


fermentfan:
@mirceanis I have a hard time finding out whether these test results are really showing an error in the system. The generated signatures mismatch from the expected ones, but maybe that is ok as the content changed with the bigint usage?

image

These are the only tests currently failing. Now waiting for release of ethr-did-resolver and ethr-did.

fermentfan:
I merge next into the branch and switched to the official releases of the 'ethr-did-resolver' & 'ethr-did'. Ready for review!

packages/kms-local/src/key-management-system.ts
```diff
@@ -257,14 +264,19 @@ export class KeyManagementSystem extends AbstractKeyManagementSystem {
    * @returns a `0x` prefixed hex string representing the signed digest in compact format
    */
   private eth_rawSign(managedKey: string, data: Uint8Array) {
-    return new SigningKey('0x' + managedKey).signDigest(data).compact
+    // Is compact serialized really what we want?
+    return new SigningKey('0x' + managedKey).sign(data).compactSerialized
```
mirceanis, line 268: `compactSerialized` in v6 seems to be the equivalent of `compact` in v5, so it should work as before.

You are right to question it as we need to document things better.

packages/kms-local/src/key-management-system.ts
```diff
@@ -257,14 +264,19 @@ export class KeyManagementSystem extends AbstractKeyManagementSystem {
...
   /**
    * @returns a base64url encoded signature for the `EdDSA` alg
    */
   private async signEdDSA(key: string, data: Uint8Array): Promise<string> {
-    const signer = EdDSASigner(arrayify(key, { allowMissingPrefix: true }))
+    //TODO: Make sure this works as we removed the option from the arrayify method
+    if (key && key.substring(0, 2) !== "0x") {
+      key = "0x" + key;
+    }
+    const signer = EdDSASigner(getBytes(key))
```
mirceanis, line 279: It's easier to use `@veramo/utils#hexToBytes`. It will take care of prefixes as well.

```suggestion
    const signer = EdDSASigner(hexToBytes(key))
```

Also true for the next few signing methods in this class

packages/kms-web3/package.json
```diff
@@ -10,12 +10,10 @@
...
   },
   "dependencies": {
-    "@ethersproject/providers": "^5.7.0",
-    "@ethersproject/strings": "^5.7.0",
-    "@ethersproject/transactions": "^5.7.0",
     "@veramo/core-types": "workspace:^",
     "@veramo/key-manager": "workspace:^",
-    "debug": "^4.3.3"
+    "debug": "^4.3.3",
+    "ethers": "6.7.1"
```
mirceanis, line 16: Please accept wider version ranges for dependencies
```suggestion
    "ethers": "^6.7.1"
```

packages/test-react-app/src/veramo/setup.ts
```diff
@@ -107,7 +107,7 @@ export function getAgent(options?: IAgentOptions): TAgent<InstalledPlugins> {
                 rpcUrl: 'https://goerli.infura.io/v3/' + INFURA_PROJECT_ID,
               },
               {
-                chainId: 421613,
+                chainId: BigInt(421613),
```
mirceanis, line 110: Please revert the use of BigInt in configs
```suggestion
                chainId: 421613,
```

mirceanis (changes requested):
Overall pretty good but some small adjustments are still needed.
Thanks for powering through!

packages/key-manager/src/abstract-key-management-system.ts
```diff
@@ -20,7 +19,8 @@ export abstract class AbstractKeyManagementSystem {
   /**@deprecated please use `sign({key, alg: 'eth_signTransaction', data: arrayify(serialize(transaction))})` instead */
   async signEthTX({ key, transaction }: { key: Pick<IKey, 'kid'>; transaction: object }): Promise<string> {
     const { v, r, s, from, ...tx } = <any>transaction
-    const data = arrayify(serialize(tx))
+    const serializedTx = Transaction.from(tx).serialized
```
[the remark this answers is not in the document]
mirceanis: perhaps this is also the source of the failed tests

packages/did-provider-ethr/src/kms-eth-signer.ts
```diff
@@ -1,25 +1,30 @@
...
-export class KmsEthereumSigner extends Signer implements TypedDataSigner {
+export class KmsEthereumSigner extends AbstractSigner {
   private context: IRequiredContext
   private controllerKey: IKey
-  readonly provider?: Provider
+  readonly provider: Provider
 
-  constructor(controllerKey: IKey, context: IRequiredContext, provider?: Provider) {
-    super()
+  constructor(controllerKey: IKey, context: IRequiredContext, provider: Provider) {
```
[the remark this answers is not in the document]
fermentfan: It's very weird. The `AbstractSigner` dictates the class' provider field to not be undefined. So it always expects one to be set in the constructor.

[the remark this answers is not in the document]
fermentfan: Found a way now to achieve this. They did some heavy weird shit with type arguments possibly being null even though the field enforced not-null and not-undefined.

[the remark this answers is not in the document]
mirceanis: cool!

mirceanis:
I figured out why the tests are failing there.

ethers infers the transaction type when serializing it. v6 does the inference a little differently and ends up serializing transactions that specify a gasPrice as type 1 where v5 would infer type 0.

Both transaction forms would be accepted by the EVM so it is ok to update the test vectors with the new values.

For completeness, we should also have a test that sets the transaction type to 0 and compares with the previous test vectors:

```ts
      it('should sign EthTX using generic signer and specific type', async () => {
        const transaction = Transaction.from({
          to: '0xcE31a19193D4b23F4E9D6163d7247243BAF801C3',
          value: 300000,
          gasLimit: 43092000,
          gasPrice: 20000000000,
          nonce: 1,
          type: 0, // enforce legacy serialization
        })

        const txData = transaction.unsignedSerialized

        const rawTx = await agent.keyManagerSign({
          algorithm: 'eth_signTransaction',
          data: txData,
          encoding: 'hex',
          keyRef: importedKey.kid,
        })

        expect(rawTx).toEqual(
          '0xf869018504a817c800840291882094ce31a19193d4b23f4e9d6163d7247243baf801c3830493e0801ba0f16e2206290181c3feaa04051dad19089105c24339dbdf0d80147b48a59fa152a0770e8751ec77ccc78e8b207023f168444f7cfb67055c55c70ef75234458a3d51',
        )
      })
```

fermentfan:
Ahh thanks for spotting that. Amazing! I pushed the changes including the new  test now. Let's see what the CI/CD says. At least on my local machine everything is green

packages/key-manager/src/abstract-key-management-system.ts
```diff
@@ -31,7 +31,11 @@ export abstract class AbstractKeyManagementSystem {
     let dataBytes: Uint8Array
     if (typeof data === 'string') {
       try {
-        dataBytes = arrayify(data, { allowMissingPrefix: true })
+        // TODO: Make sure this works as we removed the options from arrayify
+        if (data && data.substring(0, 2) !== "0x") {
+          data = "0x" + data;
+        }
+        dataBytes = getBytes(data)
```
mirceanis, line 38:
```suggestion
        dataBytes = hexToBytes(data)
```

mirceanis (approved):
Looks good.

There are plenty of changes not covered by tests, but I think the original code paths weren't covered either.

PR #1249, opened 2023-09-20
fix(data-store): take and skip for postgres

simonas-notcat:

When running `dataStoreORMGetVerifiableCredentials` with :

```
{
    "order": [
        {
            "column": "issuanceDate",
            "direction": "DESC"
        }
    ],
    "take": 10,
    "skip": 0
}
```

on an agent that is using Postgres DB you would get this error: 

```
for SELECT DISTINCT, ORDER BY expressions must appear in select list
```

## What is being changed
Using `qb.offset()` and `qb.limit()` instead of `qb.skip()` and `qb.take()`

As suggested here: https://github.com/typeorm/typeorm/issues/4742#issuecomment-783857414

## Quality
Check all that apply:
* [ ] I added unit tests.

### DIFF
--- a/packages/data-store/src/data-store-orm.ts
+++ b/packages/data-store/src/data-store-orm.ts
@@ -411,8 +411,8 @@ function decorateQB(
   tableName: string,
   input: FindArgs<any>,
 ): SelectQueryBuilder<any> {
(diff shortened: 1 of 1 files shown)


mirceanis (approved):
Looks good

PR #1282, opened 2023-10-24
feat(coordinate-mediation): upgrade from 2.0 to 3.0

### LINKED ISSUE #1255: Add Support for DIDCommv3 Coordinate Mediation
## Overview

Veramo has already implemented DIDComm v2 Coordinate Mediation. With the recently released [DIDComm v3 Specification for Coordinate Mediation](https://didcomm.org/coordinate-mediation/3.0/), the work already done in this [PR](https://github.com/uport-project/veramo/pull/1095) will need to be updated. 

## Solution

 Would you be open to contributions to transfer the existing implementation to v3?

## Resources

- [DIDComm v3 Coordinate Mediation Specification](https://didcomm.org/coordinate-mediation/3.0/)
- [Veramo DIDComm v2 Mediation Issue](https://github.com/uport-project/veramo/issues/1051)

radleylewis:
closes #1255

## What is being changed

This PR updates Veramo's Coordinate Mediation from the v2 specification to the recently released [v3 Coordinate Mediation](https://didcomm.org/coordinate-mediation/3.0/) specification. A summary of the changes introduced is as follows:

**Coordinate Mediation:**
- update Mediate Request (and its response types of Mediate Grant/Mediate Deny) 
- adds Recipient Update and Recipient Update Response
- adds Recipient Query and Recipient

**Mediate Grant or Deny control mechanism:**
- adds a Mediate Grant or Mediate Deny control mechanism
- mechanism is currently hardcoded, but TODO marked to allow injection of this as default or a user-defined custom handler
- adds a MediatePolicy data table
- adds a global rule to allow-all or deny-all by default
- adds core-types for Mediate Policy 
- adds associated dataStore methods
- adds associated declarations

**mediate cli command**
- adds the mediate cli command
- adds: list, remove, allow-from and deny-from sub-commands

**Track Granted and Denied Mediation Requests**
- adds a Mediation table to the dataStore to track GRANTED/DENIED status
- adds associated dataStore methods 
- adds associated type declarations

**Recipient Update**
- adds a recipient_dids table to record recipient dids
- adds associated dataStore methods
- adds associated type declarations

**Testing & Documentation**
- builds upon existing tests
- introduces additional tests for new functionality in coordinate mediation messages

## Quality
Check all that apply:
* [x] I successfully ran `pnpm i`, `pnpm build`, `pnpm test`, `pnpm test:browser` locally.
* [x] I added unit tests.

## Details


nickreynolds (commented):
overall this seems like a great update. I think @mirceanis has some thoughts on the changes to the data store, so I'll hold off on approval until hearing from him

packages/did-comm/src/types/IDIDComm.ts
```diff
@@ -99,6 +99,14 @@ export interface IDIDComm extends IPluginMethodMap {
    */
   sendDIDCommMessage(args: ISendDIDCommMessageArgs, context: IAgentContext<IResolver>): Promise<string>
 
+  /**
+   * @return a boolean indicating whether the default mediation policy is to grant all or deny all
```
pauldesmondparker, line 103:
```suggestion
   * @returns a boolean indicating whether the default mediation policy is to grant all or deny all
```

packages/did-comm/src/protocols/coordinate-mediation-message-handler.ts
```diff
@@ -9,77 +16,179 @@ const debug = Debug('veramo:did-comm:coordinate-mediation-message-handler')
 
 type IContext = IAgentContext<IDIDManager & IKeyManager & IDIDComm & IDataStore>
 
+const ALLOW = MediationPolicies.ALLOW
+const DENY = MediationPolicies.DENY
+
+export enum UpdateAction {
```
pauldesmondparker, line 22: Add documentation block with `@beta`

packages/did-comm/src/protocols/coordinate-mediation-message-handler.ts
```diff
@@ -9,77 +16,179 @@ const debug = Debug('veramo:did-comm:coordinate-mediation-message-handler')
...
+  ADD = 'add',
+  REMOVE = 'remove',
+}
+
+export enum RecipientUpdateResult {
```
pauldesmondparker, line 27: Add documentation block with `@beta`

packages/did-comm/src/protocols/coordinate-mediation-message-handler.ts
```diff
@@ -9,77 +16,179 @@ const debug = Debug('veramo:did-comm:coordinate-mediation-message-handler')
...
+  SUCCESS = 'success',
+  NO_CHANGE = 'no_change',
+  CLIENT_ERROR = 'client_error',
+  SERVER_ERROR = 'server_error',
+}
+
+export interface Update {
```
pauldesmondparker, line 34: Add documentation block with `@beta`

packages/did-comm/src/protocols/coordinate-mediation-message-handler.ts
```diff
@@ -9,77 +16,179 @@ const debug = Debug('veramo:did-comm:coordinate-mediation-message-handler')
...
+  recipient_did: string
+  action: UpdateAction
+}
+
+export interface UpdateResult extends Update {
```
pauldesmondparker, line 39: Add documentation block with `@beta`

packages/data-store/src/entities/mediation.ts
```diff
@@ -0,0 +1,14 @@
+import { MediationStatus } from '@veramo/core-types'
+import { Column, Entity, BaseEntity, PrimaryGeneratedColumn } from 'typeorm'
+
+@Entity('mediation')
```
pauldesmondparker, line 4: Add documentation block with at least `@beta`

packages/data-store/src/entities/mediation_policy.ts
```diff
@@ -0,0 +1,14 @@
+import { MediationPolicies } from '@veramo/core-types'
+import { Column, Entity, BaseEntity, PrimaryGeneratedColumn } from 'typeorm'
+
+@Entity('mediation_policy')
```
pauldesmondparker, line 4: Add documentation block with at least `@beta`

packages/data-store/src/entities/recipient_did.ts
```diff
@@ -0,0 +1,13 @@
+import { Entity, BaseEntity, Column, PrimaryGeneratedColumn } from 'typeorm'
+
+@Entity('recipient_did')
```
pauldesmondparker, line 3: Add documentation block with at least `@beta`

packages/message-handler/src/message.ts
```diff
@@ -35,6 +35,8 @@ export class Message implements IMessage {
   // @ts-ignore
   type: string
 
+  body?: Record<string, unknown>
```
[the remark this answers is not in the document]
radleylewis: Thanks for catching this. I will remove this as it is not used.

packages/did-comm/src/__tests__/routing-message-handler.test.ts
```diff
@@ -107,7 +108,7 @@ describe('routing-message-handler', () => {
     })
 
     recipient = await agent.didManagerImport({
-      did: 'did:fake:z6MkgbqNU4uF9NKSz5BqJQ4XKVHuQZYcUZP8pXGsJC8nTHwo',
+      did: 'did:fake:routerqNU4uF9NKSz5BqJQ4XKVHuQZYcUZP8pXGsJC8nTHwo',
```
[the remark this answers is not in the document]
radleylewis: Nice catch. I will revert back to the original.

packages/did-comm/src/didcomm.ts
```diff
@@ -177,14 +178,18 @@ export class DIDComm implements IAgentPlugin {
    * @param transports - A list of {@link IDIDCommTransport} objects. Defaults to
    *   {@link @veramo/did-comm#DIDCommHttpTransport | DIDCommHttpTransport}
    */
-  constructor(transports: IDIDCommTransport[] = [new DIDCommHttpTransport()]) {
+  constructor(
+    transports: IDIDCommTransport[] = [new DIDCommHttpTransport()],
+    isMediateDefaultGrantAll = true,
+  ) {
```
pauldesmondparker, line 184: Given that `DIDComm` is still marked `@beta` it might be better, given multiple optional parameters with defaults to use an object parameter:
```suggestion
  constructor({
    transports = [new DIDCommHttpTransport()],
    isMediateDefaultGrantAll = true
  }: { transports?: IDIDCommTransport[], isMediateDefaultGrantAll?: boolean }) {
```

packages/did-comm/src/__tests__/coordinate-mediation-message-handler.test.ts
```diff
@@ -83,11 +90,9 @@ describe('coordinate-mediation-message-handler', () => {
             ...new FakeDidResolver(() => agent).getDidFakeResolver(),
           }),
         }),
-        // @ts-ignore
         new DIDComm([new DIDCommHttpTransport()]),
```
pauldesmondparker, line 93: As consequence to changing parameter of DIDComm to an object.
```suggestion
        new DIDComm({ transports: [new DIDCommHttpTransport()]}),
```

packages/did-comm/src/didcomm.ts
pauldesmondparker: This affects the following code locations:
```grep
packages/did-comm/src/__tests__/message-handler.test.ts:        new DIDComm([new DIDCommHttpTransport()]),
packages/did-comm/src/__tests__/trust-ping-message-handler.test.ts:        new DIDComm([new DIDCommHttpTransport()]),
packages/did-comm/src/__tests__/routing-message-handler.test.ts:        new DIDComm([new DIDCommHttpTransport()]),
packages/did-comm/src/__tests__/messagepickup-message-handler.test.ts:          new DIDComm([new DIDCommHttpTransport()]),
packages/did-comm/src/__tests__/messagepickup-message-handler.test.ts:          new DIDComm([new DIDCommHttpTransport()]),
__tests__/localAgent.test.ts:      new DIDComm([new DIDCommHttpTransport()]),
__tests__/restAgent.test.ts:      new DIDComm([new DIDCommHttpTransport()]),
```
Where:
```ts
new DIDComm([new DIDCommHttpTransport()])
```
becomes:
```ts
new DIDComm({ transports: [new DIDCommHttpTransport()]})
```

pauldesmondparker: It might also be a good idea to add tests for the `isMediateDefaultGrantAll` parameter.

radleylewis: I agree this is a cleaner approach. In the `agent.yml` file the syntax would be as follows:

```yaml
- $require: '@veramo/did-comm#DIDComm'
  $args:
    - isDefaultMediateGrantAll: false
      transports:
        $require: '@veramo/did-comm#DIDCommHttpTransport'
```

I have included this and pushed the update. NOTE the implementation of one minor adjustment which is to set the default constructor argument to an empty object allowing for the agent.yml file to exclude these args entirely and still invoke the default configuration as per your implemetation (I checked and without $args the parameter will be undefined).

Updated the invocation of DIDComm with the new argument parameters.

packages/did-comm/src/__tests__/coordinate-mediation-message-handler.test.ts
radleylewis: Cheers, updated (using your patch)

pauldesmondparker: #resolved

packages/data-store/src/data-store.ts
```diff
@@ -142,4 +170,81 @@ export class DataStore implements IAgentPlugin {
...
+    did,
+  }: IDataStoreRemoveMediationPolicyArgs): Promise<RemoveMediationPolicyResult> {
+    const db = await getConnectedDb(this.dbConnection)
+    const existingEntry = await db.getRepository(MediationPolicy).findOne({ where: { did } })
+    if (!existingEntry) return null
+    await db.getRepository(MediationPolicy).remove(existingEntry)
+    return existingEntry.did
+  }
+
+  async dataStoreGetMediationPolicies({
```
pauldesmondparker, line 192:
```suggestion
/**
 * This is used internally when global policy `isMediateDefaultGrantAll` is:
 *  - true: with policy = MediatePolicy.DENY [All DIDs allowed except those explicitly denied]
 *  - false: with policy = MediatePolicy.ALLOW [All DIDs denied except those explicitly allowed]
 *
 * @beta This API may change in future versions without a BREAKING CHANGE notice.
 */
  async dataStoreGetMediationPolicies({
```

radleylewis: Thanks for this. Good suggestion, I have included it in the PR.

packages/message-handler/src/abstract-message-handler.ts
```diff
@@ -18,9 +18,7 @@ export abstract class AbstractMessageHandler {
   }
 
   public async handle(message: Message, context: IAgentContext<{}>): Promise<Message> {
-    if (this.nextMessageHandler) {
-      return this.nextMessageHandler.handle(message, context)
-    }
+    if (this.nextMessageHandler) return this.nextMessageHandler.handle(message, context)
```
mirceanis, line 21: Please revert this
```suggestion
    if (this.nextMessageHandler) {
      return this.nextMessageHandler.handle(message, context)
    }
```

packages/did-comm/src/didcomm.ts
```diff
@@ -177,14 +188,18 @@ export class DIDComm implements IAgentPlugin {
    * @param transports - A list of {@link IDIDCommTransport} objects. Defaults to
    *   {@link @veramo/did-comm#DIDCommHttpTransport | DIDCommHttpTransport}
    */
-  constructor(transports: IDIDCommTransport[] = [new DIDCommHttpTransport()]) {
+  constructor({
+    transports = [new DIDCommHttpTransport()],
+    isMediateDefaultGrantAll = true,
+  }: DIDCOmmConfig = {}) {
```
mirceanis, line 194:
```suggestion
  }: DIDCommConfig = {}) {
```

packages/did-comm/src/protocols/coordinate-mediation-message-handler.ts
mirceanis, line 1: It looks like the v2 behavior is being replaced which would make this a breaking change.
Ideally the v2 should be marked as deprecated and a v3 protocol added as a replacement.

mirceanis (changes requested):
Thanks for this awesome contribution, it's really great to see such a high quality PR as a first contribution!

There are a couple of things to be addressed before we can merge this:
* There's a lot of coupling between the new protocol and the data-store API + implementations which will definitely cause lots of friction down the road. It's best to separate the config storage into its own plugin.
* The v2 functionality is being replaced altogether. Anyone using or depending on v2 will be stuck on an old version of veramo if we merge it as is.

packages/did-comm/src/protocols/coordinate-mediation-message-handler.ts
pauldesmondparker: We considered this, and chose to replace v2 with v3 because it was marked `@Beta`, indicating that users should expect it to break without warning. Ref: https://api-extractor.com/pages/tsdoc/doc_comment_syntax/#release-tags
image

If we are to retain V2, should we mark V2 and V3 as `@public` ?

packages/message-handler/src/abstract-message-handler.ts
pauldesmondparker: We can revert this, but we have a suggestion.

The project is currently setup to use `prettier`:
```json
    "prettier": "prettier --write \"{packages,docs,__tests__,!build}/**/*.{ts,js,json,md,yml}\"",
```
And `prettier` is opinionated about single lines inside an `if` block:
image

> [!Note]
> This script has not been required for PRs and many files are currently violating the `prettier` formatting rules. You could add this to your CI to automatically catch when contributors fail to run `pnpm prettier` before submission.

packages/did-comm/src/protocols/coordinate-mediation-message-handler.ts
pauldesmondparker: @mirceanis My bad.  I made the call to replace V2 with V3, and I've realized my mistake after reading your comment.  V2 can be considered a completely separate implementation, that may continue to change over time, but it'll always be V2.

radleylewis: OK, this makes sense. I will restore the existing v2 Coordinate Mediation implementation protocol and shift the v3 implementation introduced by this PR into a new protocol file. 

Just one thing to consider is the `routing-message-handler` which also had to be changed in order to accommodate the v3 specification flow. Given we are introducing the `kv-store` on v3, but want to avoid breaking changes, I will need to ensure both cases are dealt with (v2 and v3) here. If you have a specific idea on how best to implement this I am keen to hear it, but otherwise this is just a heads up on the separation of concerns with respect to v2 and v3.

packages/data-store-json/src/data-store-json.ts
```diff
@@ -86,7 +119,15 @@ export class DataStoreJson implements IAgentPlugin {
...
-      //dataStoreDeleteVerifiablePresentation: this.dataStoreDeleteVerifiablePresentation.bind(this),
+      dataStoreSaveMediationPolicy: this.dataStoreSaveMediationPolicy.bind(this),
+      dataStoreRemoveMediationPolicy: this.dataStoreRemoveMediationPolicy.bind(this),
+      dataStoreGetMediationPolicies: this.dataStoreGetMediationPolicies.bind(this),
+      dataStoreIsMediationGranted: this.dataStoreIsMediationGranted.bind(this),
+      dataStoreSaveMediation: this.dataStoreSaveMediation.bind(this),
+      dataStoreGetMediation: this.dataStoreGetMediation.bind(this),
+      dataStoreAddRecipientDid: this.dataStoreAddRecipientDid.bind(this),
+      dataStoreRemoveRecipientDid: this.dataStoreRemoveRecipientDid.bind(this),
+      dataStoreGetRecipientDids: this.dataStoreGetRecipientDids.bind(this),
```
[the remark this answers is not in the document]
radleylewis: Thanks for this feedback. The dataStore changes definitely feel clunky in their current form and so I appreciate the guidance here and what you have outlined makes sense. 

With respect to implementing the `kv-store`, I couldn't see any current examples, so just wanted to outline my understanding here so it can be corrected early. 

**1. Implementation of the `kv-store`**
The `kv-store` should be configured in the agent.yml

**2. a new plugin should be created (e.g. `CoordinateMediationPolicy` or similar) **
The new plugin will require the `kv-store` and then be available on the agent (so shift all of the changes from the dataStore)

**3. Mediators operating on the V3 Coordinate Mediation specification install the plugin**
This means that the newly introduced v3 coordinate mediation message handler protocol implementation and the updated routing message handler protocol can access the `kv-store` with specific methods (such as the one you outlined above `coordinateMediationV3SavePolicy`). 

Thanks again for the guidance and I will check in on any feedback.

packages/did-comm/src/protocols/coordinate-mediation-message-handler.ts
pauldesmondparker: To be clear, `routing-message-handler` changed because we introduced a default mechanism for handling the grant or deny mediation choice based on the requesting DID, and a way of mutating these grant or deny records from the command line and/or API.

`routing-massage-handler` (routing is V2, and no v3 exists) will need to handle the old method of relying on previously routed messages to decide if a message should be routed, and the new of querying the `kv-store` for `denyFrom` or `grantFrom` records.

We intend to instantiate the `kv-store` on agent when using V3, and if it doesn't exist, use the previous behavior. There's a corner case here if the `kv-store` is added to the agent for another reason, but the old behavior is desired, i.e. not using `coordinate-mediation` v3.

Another option is to add an optional argument to the `routing-message-handler` to explicitly select the behavior, defaulting to the old behavior, this wouldn't break existing v2 usage.

Could you also comment on the appropriateness of putting `ks-store` on agent?  We have not found any exemplars of `kv-store` usage as it relates to the `agent.yml` file.

packages/did-comm/src/didcomm.ts
```diff
@@ -141,6 +142,16 @@ export interface IPackDIDCommMessageArgs {
...
 }
 
+/**
+ * The config for the {@link DIDComm} DIDComm plugin.
+ *
+ * @beta This API may change without a BREAKING CHANGE notice.
+ */
+export interface DIDCOmmConfig<T extends IDIDCommTransport = DIDCommHttpTransport> {
+  transports?: T[]
+  isMediateDefaultGrantAll?: boolean
```
pauldesmondparker, line 152: @radleylewis You know this already, but I'll jot it down here. Move this parameter to the `CoordinateMediationMediatorMessageHandler` constructor. You'll need to put the `grantOrDenyMediation` inside the `CoordinateMediationMediatorMessageHandler` class, so it has access to the argument.

radleylewis: Sounds good.

radleylewis: This has  been shifted to the `MediationManagerPlugin`.

packages/data-store-json/src/data-store-json.ts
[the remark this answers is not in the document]
radleylewis: Hi @mirceanis. Thanks again for the feedback. 

Just as an update, the following changes have been implemented in line with what has been discussed:

**1. `kv-store`**
The `kv-store` is now used for Mediation related storage with the use of the KeyValueTypeORMAdapter for persistent storage. This means that all previous changes to the `data-store` have been rolled back and are now encapsulated in the plugin (described below). 

**2. MediationManagerPlugin**
The PR now introduces a `MediationManagerPlugin` which allows us to expose methods pertaining to coordinate mediation on the agent. These methods are essentially ported like-for-like from what had been implemented on the `data-store` previously. 

**NOTE:**

A note on the `kv-store`. At present, there is no way to read out all of the keys in the `kv-store` for a given namespace. This is an issue because RECIPIENT_QUERY requires us to be able to list the `recipient_did`s. This constraint means that I have disabled the `cli` `mediate` `listPolicies`, and, more importantly, disabled the tests on coordinate medation v3 for RECIPIENT_QUERY. Do you have a proposal with respect to this? We could consider expanding the methods of the `kv-store` to allow for such a method to be exposed/adjust the existing `getMany` but first want to fully understand any considerations that you may have in mind. 

I look forward to your feedback. Thanks again.

packages/did-comm/src/protocols/coordinate-mediation-message-handler.ts
radleylewis: I have allowed for backward compatibility with v2 coordinate mediation in the `routing-message-handler`. I have also included tests for both. The determination as to whether the  v2 or v3 mediation will be used is through a check on whether the `isDefaultMediateGrantAll` method exists on the agent. This is necessary because all of the methods are exposed directly on the agent (so checking if the `MediationManagerPlugin` is available isn't possible, however, checking if this method (mentioned above) exists, indicates whether the plugin is loaded. 

Alternatively, we could have an optional argument, as outlined by @pauldesmondparker.

pauldesmondparker:
@mirceanis Just a reminder that this is here :v:

mirceanis:
> @mirceanis Just a reminder that this is here ✌️

I saw it, but I'm traveling this week so I'll be able to review only next week. Thank you for the effort and the patience.

radleylewis:
> closes #1255
> 
> ## What is being changed
> This PR updates Veramo's Coordinate Mediation from the v2 specification to the recently released [v3 Coordinate Mediation](https://didcomm.org/coordinate-mediation/3.0/) specification. A summary of the changes introduced is as follows:
> 
> **Coordinate Mediation:**
> 
> * update Mediate Request (and its response types of Mediate Grant/Mediate Deny)
> * adds Recipient Update and Recipient Update Response
> * adds Recipient Query and Recipient
> 
> **Mediate Grant or Deny control mechanism:**
> 
> * adds a Mediate Grant or Mediate Deny control mechanism
> * mechanism is currently hardcoded, but TODO marked to allow injection of this as default or a user-defined custom handler
> * adds a MediatePolicy data table
> * adds a global rule to allow-all or deny-all by default
> * adds core-types for Mediate Policy
> * adds associated dataStore methods
> * adds associated declarations
> 
> **mediate cli command**
> 
> * adds the mediate cli command
> * adds: list, remove, allow-from and deny-from sub-commands
> 
> **Track Granted and Denied Mediation Requests**
> 
> * adds a Mediation table to the dataStore to track GRANTED/DENIED status
> * adds associated dataStore methods
> * adds associated type declarations
> 
> **Recipient Update**
> 
> * adds a recipient_dids table to record recipient dids
> * adds associated dataStore methods
> * adds associated type declarations
> 
> **Testing & Documentation**
> 
> * builds upon existing tests
> * introduces additional tests for new functionality in coordinate mediation messages
> 
> ## Quality
> Check all that apply:
> 
> * [x]  I want these changes to be integrated
> * [x]  I successfully ran `pnpm i`, `pnpm build`, `pnpm test`, `pnpm test:browser` locally.
> * [x]  I allow my PR to be updated by the reviewers (to speed up the review process).
> * [x]  I added unit tests.
> * [x]  I added integration tests.
> * [ ]  I did not add automated tests because _________, and I am aware that a PR without tests will likely get rejected.
> 
> ## Details
> If applicable, add screen captures, error messages or stack traces to help explain your problem.

To make it easier to track the scope of the PR I have summarised the main changes introduced by this PR below:

* retain coordinate-mediation v2 (protocol)
* add coordinate-mediation v3 (protocol)
* mediation manager (plugin):
  - add mechanism to choose between all deny OR all accept mediation strategies
  - add cli and api support for adding specific mediation accept or deny decisions in advance of mediation requests
  - use kv-store as storage adapter for mediation plugin

packages/mediation-manager/README.md
```diff
@@ -0,0 +1,124 @@
...
+    - type: sqlite
+      database:
+        $ref: /constants/databaseFile
+      synchronize: false
+      migrationsRun: true
+      migrations:
+        $require: '@veramo/data-store?t=function#migrationConcat'
+        $args:
+          - $require: '@veramo/data-store?t=object#migrations'
+          - $require: '@veramo/kv-store?t=object#kvStoreMigrations'
```
mirceanis, line 27: neat solution for combining migrations!

packages/mediation-manager/src/mediation-manager.ts
```diff
@@ -0,0 +1,130 @@
...
+  readonly #preRequestPolicyStore: KeyValueStore<PreMediationRequestPolicy>
+  readonly #mediationResponseStore: KeyValueStore<MediationResponse>
+  readonly #recipientDidStore: KeyValueStore<RequesterDid>
+  readonly methods: IMediationManager
+
+  constructor(
+    isMediateDefaultGrantAll = true,
+    preRequestPolicyStore: PreRequestPolicyStore,
+    mediationResponseStore: MediationResponseStore,
+    recipientDidStore: RecipientDidStore,
```
mirceanis, line 41: Besides the data type of the value, can you think of situations when these kv-stores would need to be different?
Would you object to merging these into a single kv-store later on?

packages/message-handler/src/abstract-message-handler.ts
mirceanis: I agree. We have to re-enable automatic formatting.
We've been putting it off because of pending PRs that would show conflicts, but you are 100% right it should be done.

packages/mediation-manager/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2019 Consensys AG
```
mirceanis, line 189: This is not accurate for new code, and since this codebase now sits in DIF I'm not sure what the correct value should be here.. maybe just your name?

mirceanis (approved):
It looks much better now with an independent plugin for managing mediation metadata.
Kudos for the various workarounds you've found! It's been a delight reviewing this :)

This PR can probably be merged as is.
But if you still have time for some improvements, I have a few points:
* revert the did-comm plugin constructor changes
* define the `IMediationManager` and collateral types in the new package instead of the `@veramo/core-types`
* maybe simplify the mediation manager plugin config by using a single kv-store?

packages/cli/src/mediate.ts
```diff
@@ -0,0 +1,178 @@
...
+}
+
+async function listPolicies(options: Pick<Options, 'allowFrom' | 'denyFrom'>, cmd: Command): Promise<void> {
+  try {
+    // NOTE: disabled as kv-store has no getAll method
+    // const agent = await getAgent(cmd.optsWithGlobals().config)
+    // const policies = await agent.dataStoreGetMediationPolicies()
+    // if (options.allowFrom) return console.log(policies.filter((policy) => policy.policy === ALLOW))
+    // else if (options.denyFrom) return console.log(policies.filter((policy) => policy.policy === DENY))
+    // else console.log(policies)
```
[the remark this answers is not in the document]
pauldesmondparker: This is our backstop work-around as well. If we can't change the behavior of `kv-store`, then this is what we'll go with.

The trade-off is a certain amount of coding complication (plus trivial increase in computation) for CRUD on list values.
|                  | Create    | Read      | Read All   | Update    | Delete    |
|------------------|-----------|-----------|------------|-----------|-----------|
| key-value (current)        | Easy      | Easy      | Impossible | Easy      | Easy      |
| "list": values[] | Expensive | Expensive | Easy       | Expensive | Expensive |

packages/mediation-manager/README.md
radleylewis: Credit goes to @pauldesmondparker for this one!

packages/mediation-manager/LICENSE
radleylewis: Good catch. The `mediation-manager` plugin was inspired by the template hence the oversight. I have updated it with my own name.

packages/mediation-manager/src/mediation-manager.ts
pauldesmondparker: I feel the trade-off here is between ease of setup and code readability. There is also intersection with the decision to modify or not modify the behavior of `kv-store` itself.
|            | agent setup | code clarity | key prefixes |
|------------|-------------|--------------|--------------|
| pre-policy | Verbose     | Good         | None         |
| response   | Verbose     | Good         | None         |
| recipients | Verbose     | Good         | None         |
| combined   | Concise     | Less Clear   | 3 prefixes   |

> [!Note]
> The context here is if `kv-store` were modified to allow `list-all` functionality. Otherwise the prefixes would just be the current `namespace` + "-list" and reference a stringified JSON value each.
> _____
> pre-policy key prefix: `pre-` -> `pre-{requester-did}`
> response key prefix: `response-` -> `response-{requester-did}`
> recipients key prefix: `recipient-` -> `recipient-{recipient-did}`

> [!Warning]
> Question 1: is the proliferation of `kv-store` namespaces (as an isolated factor) any issue?
> Question 2: is the reduced verbosity of `agent.yml` worth the complication?
> Question 3: is there a strong resistance to adding some kind of `list-all` functionality to `kv-store`? (we have multiple ways of achieving this, one that changes the underlying `keyv` and one that doesn't).

> [!Note]
> My two cents is that if we use lists of values, rather than pure `key-value` pairs, then it makes sense to combine everything into one namespace, since we will essentially just have three large rows.

package.json
```diff
@@ -13,6 +13,7 @@
     "test:integration": "pnpm test:integration-build && pnpm test:ci",
     "test:ci": "pnpm test -- --coverage=true",
     "test": "cross-env NODE_OPTIONS=\"--experimental-vm-modules --experimental-import-meta-resolve\" jest",
+    "test:did-comm": "cross-env NODE_OPTIONS=\"--experimental-vm-modules --experimental-import-meta-resolve\" jest did-comm",
```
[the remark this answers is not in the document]
radleylewis: Removed.

packages/mediation-manager/src/mediation-manager.ts
```diff
@@ -0,0 +1,130 @@
...
+ *
+ * @remarks be advised that the {@link @veramo/core-types#IMediationManager} interface is for use with
+ * {@link @veramo/did-comm#DIDComm | DIDCOmm} and specifically the V3 Coordinate Mediation Protocol implementation.
+ *
+ * @beta This API may change without a BREAKING CHANGE notice.
+ */
+export class MediationManagerPlugin implements IAgentPlugin {
+  readonly #preRequestPolicyStore: KeyValueStore<PreMediationRequestPolicy>
+  readonly #mediationResponseStore: KeyValueStore<MediationResponse>
+  readonly #recipientDidStore: KeyValueStore<RequesterDid>
```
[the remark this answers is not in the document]
radleylewis: No worries. I have removed the JS `#` and replaced it with the `private` TS keyword.

radleylewis:
> It looks much better now with an independent plugin for managing mediation metadata. Kudos for the various workarounds you've found! It's been a delight reviewing this :)
> 
> This PR can probably be merged as is. But if you still have time for some improvements, I have a few points:
> 
> * revert the did-comm plugin constructor changes
> * define the `IMediationManager` and collateral types in the new package instead of the `@veramo/core-types`
> * maybe simplify the mediation manager plugin config by using a single kv-store?

Thanks for the thorough review, it is greatly appreciated! I have pushed some updates in line with your comments. Specifically:

* On the first point around _revert the did-comm plugin constructor changes_ my thinking here is that a object allowing for key parameters is more adaptable to future changes than ordered parameter arguments. I am happy to revert if there is a compelling case though. 
* The `mediation-manager` types have been shifted from the core-types to the mediation manager plugin directory. 
* The kv-store is covered in the PR comment responses. Just to point out that in some cases the keys may be the same, resulting in clashes, and so would need to be prefixed with group identifiers which does have some trade offs. 

Thanks again and we look forward to your comments.

PR #1283, opened 2023-10-25
feat(did-comm): sendDIDCommMessage - returnMessage

simonas-notcat:
## What is being changed
A return message will be included in the response to `sendDIDCommMessage` when `return_route: 'all'` is used in a packed message

```typescript
const messageId = v4()
const packedMessage = await agent.packDIDCommMessage({ 
  message: {
    type: 'example',
    from: senderDid,
    to: recipientDidUrl,
    id: messageId,
    body: {},
    return_route: 'all'
  },
  packing: 'authcrypt'
})
const { returnMessage, transportId } = await agent.sendDIDCommMessage({ 
  packedMessage, messageId, recipientDidUrl
})

console.log(returnMessage)
```

## Quality
Check all that apply:
* [X] I added unit tests.

### DIFF
--- a/packages/did-comm/src/__tests__/messagepickup-message-handler.test.ts
+++ b/packages/did-comm/src/__tests__/messagepickup-message-handler.test.ts
@@ -334,6 +334,33 @@ describe('messagepickup-message-handler', () => {
       )
     })
 
+    it('should contain returnMessage', async () => {
+
+      // Send StatusRequest
+      const statusRequestMessage: IDIDCommMessage = {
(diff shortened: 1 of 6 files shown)


mirceanis (changes requested):
There's a tiny change that needs to be made to the inline doc of the `sendDIDCommMessage` method to align with the changes.

Since the API of the method is changing we'll need to mark this as a breaking change and include it in the next major release.

mirceanis (approved):

PR #1299, opened 2023-12-08
fix(data-store-orm): remove double entries

### LINKED ISSUE #1285: remote agents returns VCs three times
**Bug severity**
5

**Describe the bug**
When using the remote agent plugin, each VC is included three times in the http response:

image

image

**To Reproduce**
1 include the remote and client plugin
2 write a credential to the database
3 request it via the remote client

When I fetch the credentials directly in the backend code, I get two credentials back. So the problem is not in the `dataStoreORMGetVerifiableCredentials`, but in the usage when used via the remote call.

**Observed behaviour**
Results do not match. I haven't checked if other functions behave the same or if it's only the dataStoreORMGetVerifiableCredentials call.
   
**Expected behaviour**
That the remote call returns the same result as the "normal" call.

**Details**
If applicable, add screenshots, error messages or stack traces to help explain your problem.

**Additional context**

 - Veramo: 5.x

cre8:
Signed-off-by: Mirko Mollik <mirko.mollik@fit.fraunhofer.de>

fixes #1285

## What is being changed
It will reduce the credentials in case there were multiuple claims pointing to the claim credential

## Quality
Check all that apply:
* [ ] I added unit tests.
* [X] I did not add automated tests because there was no existing test for this function and I am aware that a PR without tests will likely get rejected.

### DIFF
--- a/packages/data-store/src/__tests__/data-store-orm.test.ts
+++ b/packages/data-store/src/__tests__/data-store-orm.test.ts
@@ -264,7 +264,7 @@ describe('@veramo/data-store queries', () => {
     expect(credentials.length).toBe(1)
     expect(credentials[0].verifiableCredential.id).toBe('vc1')
     const count = await makeAgent().dataStoreORMGetVerifiableCredentialsByClaimsCount({})
-    expect(count).toBe(1)
+    expect(count).toBe(3)
 
     const credentials2 = await makeAgent({
       authorizedDID: did3,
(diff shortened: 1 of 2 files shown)


mirceanis (changes requested):
looks like there is a test that was validating the old buggy behavior: https://github.com/decentralized-identity/veramo/blob/85be779231f615961b259cda8229ccb3525e6d7e/packages/data-store/src/__tests__/data-store-orm.test.ts#L264

It should expect the result length to be 1 instead of 3

The same test exists in the data-store-json package and it is correctly expecting 1 there. It's a good thing you caught this bug :)

cre8:
@mirceanis didn't find a way to run the test locally just for one package. Fixed it and it passed in the ci

mirceanis (approved):
Looks great!

PR #1314, opened 2024-01-16
[proposal] Make headers call for AgentRestClient dynamic

### LINKED ISSUE #1313: [proposal] Make headers call for AgentRestClient dynamic
I am using keycloak for the authentication flow. Since the access token is only valid for some minutes, it has to be refreshed from time to time. Right now I can only pass a static token to it like:
```
        new AgentRestClient({
          url: `${environment.agent}/agent`,
          enabledMethods: enabledIssuerMethods,
          headers: {
            Authorization: `Bearer ${token}`,
          },
        }),
```

Creating an agent on demand could break some of my veramo plugins. So I thought it would be great to pass the headers either as a static value or to pass a function call.

It should be possible to pass either a `Record<string, string>` or a `(() => Record<string, string>)` for the header. It's better to make the whole header variable dynamic and not just one specific entry of the record.

Creating an agent from time to time seems not like a best practice. Sphereon updated their client sdks so it should not be a big update. Since a static object can be passed it will not break any existing implementations.

cre8:
Fixes #1313

## What is being changed
The `AgentRestClient` constructor is able to handle functions calls for the headers. This allows to pass dynamic values like an access token that needs to be updated frequently.

## Quality
Check all that apply:
```
 FAIL  packages/data-store/src/__tests__/data-store-orm.test.ts (25.45 s)

                                                                                                                                           
  ● Test suite failed to run                                                                                                               
                                                                                                                                           
    EBUSY: resource busy or locked, unlink './tmp/test-db2.sqlite'                                                                         
                                                                                                                                           
      143 |   afterAll(async () => {                                                                                                       
      144 |     ;(await dbConnection).close()                                                                                              
    > 145 |     fs.unlinkSync(databaseFile)                                                                                                
          |        ^                                                                                                                       
      146 |   })                                                                                                                           
      147 |                                                                                                                                
      148 |   test('search presentations by verifier', async () => {                                                                       
                                                                                                                                           
      at Object.<anonymous> (packages/data-store/src/__tests__/data-store-orm.test.ts:145:8)  
```
This error has nothing to do with the changes, but resulted in a incomplete test run (OS: Windows 11)


* [ ] I added unit tests.
* [X] I did not add automated tests because the package has no test implemented, and I am aware that a PR without tests will likely get rejected.


packages/remote-client/src/client.ts
```diff
@@ -20,15 +20,17 @@ export class AgentRestClient implements IAgentPlugin {
     url: string
     enabledMethods: string[]
     schema?: IAgentPluginSchema
-    headers?: Record<string, string>
+    headers?: Record<string, string> | (() => Record<string, string>)
```
mirceanis, line 23: If this function is or can be an async call then it should be signaled in the return type:
```suggestion
    headers?: Record<string, string> | (() => Promise<Record<string, string>>)
```

mirceanis (changes requested):
Thanks for posting this and for leading with an issue explaining the changes!

Just one nit to fix, otherwise looks great!

mirceanis (approved):

PR #1315, opened 2024-01-17
Remove assert import for the plugin schema in favour of bundlers

### LINKED ISSUE #1254: Remove assert import in favour of webpack
**Bug severity**
4

**Describe the bug**
When integrating some basic plugins for veramo, I get the following error:
```
./node_modules/@veramo/message-handler/build/message-handler.js - Error: Module build failed (from ./node_modules/@angular-devkit/build-angular/src/tools/babel/webpack-loader.js):
SyntaxError: /home/mimo/Projects/ssi/node_modules/@veramo/message-handler/build/message-handler.js: Support for the experimental syntax 'importAttributes' isn't currently enabled (2:66):

  1 | import { CoreEvents, } from '@veramo/core-types';
> 2 | import schema from '@veramo/core-types/build/plugin.schema.json' assert { type: 'json' };
    |                                                                  ^
  3 | import { Message } from './message.js';
  4 | import Debug from 'debug';
  5 | const debug = Debug('veramo:message-handler');

Add @babel/plugin-syntax-import-attributes (https://github.com/babel/babel/tree/main/packages/babel-plugin-syntax-import-attributes) to the 'plugins' section of your Babel config to enable parsing.
```
Since I am using Angular that is based on webpack I found no way to deal with this problem. Adding a custom webpack and setting the babel loader with the mentioned plugin does not seem to work:

```
module.exports = {
    module: {
        rules: [
          {
            test: /\.m?js$/,
            exclude: /node_modules\/(?!(@veramo)\/).*/,
            use: {
              loader: 'babel-loader',
              options: {
                presets: ['@babel/preset-react', '@babel/preset-env'],
                plugins: ['@babel/plugin-transform-runtime'],
              },
            },
          },
        ],
      },    
}
```

When I manually removed the ` assert { type: 'json' };` from all the files it ran without the errors and I could not see any side effects.

Right now I am using a script that runs after the installation of all packages and removes the assertions:
```
const fs = require('fs');
const path = require('path');

// Recursive function to process files in directories and subdirectories
function processDirectory(directory) {
    const files = fs.readdirSync(directory);

    for (const file of files) {
        const filePath = path.join(directory, file);
        const stats = fs.statSync(filePath);

        if (stats.isDirectory()) {
            processDirectory(filePath);  // Recursive call
        } else {
            processFile(filePath);
        }
    }
}

function processFile(filePath) {
    let content = fs.readFileSync(filePath, 'utf8');
    const pattern = /assert { type: 'json' };/g;

    if (pattern.test(content)) {
        console.log(`Processing file: ${filePath}`);
        content = content.replace(pattern, '');
        fs.writeFileSync(filePath, content, 'utf8');
    }
}

const rootDirectory = path.join('node_modules', '@veramo');  // Assuming the script is executed at the parent directory of @veramo
processDirectory(rootDirectory);
```

So my question is can we remove the assert feature in favour to be webpack compatible?

**To Reproduce**
Add the veramo agent to an angular application with the `did-jwt-vc` plugin

**Observed behaviour**
Having this feature in the core libraries makes it impossible to use them in non babel environments.

cre8:

Fixes #1254

## What is being changed
As described in the issue I replaced the `assert { type: 'json' }` with a classic typescript import. This allows other bundlers like webpack to deal with this without throwing an error since the assert command is not fully supported by typescript yet

I also stored the plugin.json file as non formatted, saving around 50% of the size.

There are still some assert imports in the `ld-default-context` package, this can be handled in another issue since it has nothing to do with this task of generating the schema of a plugin.

The `plugin.schema.ts` is generated in parallel to the `plugin.schema.json` so it does not break other project that are depending on them.

## Quality
Check all that apply:
* [ ] I added unit tests.
* [X] I did not add automated tests because no new functionality was added, and I am aware that a PR without tests will likely get rejected.


cre8:
Putting this "on hold" since the migration of angular to esbuild can deal with these imports. I would investigate a bit more and if it's working the PR could be closed.

cre8:
Btw it's not only a webpack problem, but also for esbuild and babel. For babel there is a plugin that can deal with `plugin-syntax-import-assertions`, but the other bundlers are lacking behind.

mirceanis (changes requested):
This looks good to me.
Since these are generated files, they will get updated as the plugin schemas change (even when updating inline docs) so it will lead to ugly diffs and conflicts.
I think we shouldn't care about having the outputs minified (or in our case not formatted) since that's a task for bundlers.

Also, this will get merged into `next` which is scheduled for a release soon including other breaking changes that will trigger a major version bump so I would opt for removing the JSON files completely

cre8:
So should be remove the json files from this PR and merge it when it fits for the major version?

mirceanis:
> So should we remove the json files from this PR and merge it when it fits for the major version?

Yes, please remove them.

We don't have an ability to mark them as deprecated, but if anyone depends on them they'll see a broken build after upgrading to the new major version and they'll be able to roll back their upgrade.

cre8:
Done

mirceanis (approved):
This looks good.
Thank you for posting these changes.

mirceanis:
I agree. Please post a PR if you're already testing this out

mirceanis:
> I agree. Please post a PR if you're already testing this out

@cre8 nevermind, I've managed to create a PR (#1327) as it was blocking another task I'm working on.
Please take a look.

PR #1318, opened 2024-01-19
[proposal] make path output from generate schema dynamic

### LINKED ISSUE #1317: [proposal] make path output from generate schema dynamic
Right now the outcoming schem.json file is static: https://github.com/decentralized-identity/veramo/blob/next/packages/cli/src/dev.ts#L187
This makes it difficult when I run the command inside a monorepo like nx. The veramo cli process is running in the root folder and the path to the api-extractor.json can be set, but not the output.

Add a new parameter that can be passed to the cli. The default one will be the current one to not break anything.

Something like `cd libs/vc-store && npx @veramo/cli -- dev generate-plugin-schema` is not good since an already installed dependency is not used.

**Additional context**

cre8:

Example:
fixes #1317

## What is being changed
Adding a new parameter to the cli so the path and file of the `plugin.schema.json`.

## Quality
Check all that apply:
* [ ] I added unit tests.
* [X] I did not add automated tests because _________, and I am aware that a PR without tests will likely get rejected.

### DIFF
--- a/packages/cli/src/dev.ts
+++ b/packages/cli/src/dev.ts
@@ -95,9 +95,11 @@ dev
     'package.json file containing a Veramo plugin interface config',
     './package.json',
   )
+  .option('-o, --output <string>', 'Output file of the schema', './src/plugin.schema.json')
 
   .action(async (options) => {
     const apiExtractorJsonPath: string = resolve(options.extractorConfig)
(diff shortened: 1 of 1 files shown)


mirceanis (approved):
looks great!

PR #1328, opened 2024-01-26
WIP: Feat/esbuild

### LINKED ISSUE #1321: [proposal] migrate to esbuild to support esm and commonjs build

cre8:

fixes #1321 

Compared to [this PR](https://github.com/decentralized-identity/veramo/pull/1322) this is a more cleaner approach to show the current status.

Current problem with ESBUILD:
- the build:js step passes for all projects without any problems. But when running the api and schemas it fails. Giving errors like:
```
> @veramo/core-types@5.6.0 generate-plugin-schema C:\Users\mollik\Projects\veramo\packages\core-types
> node ../cli/bin/veramo.js dev generate-plugin-schema

file:///C:/Users/mollik/Projects/veramo/packages/core-types/build/types/ICredentialStatusManager.js:1
import { DIDDocument, DIDResolutionOptions, DIDResolutionResult } from "did-resolver";
         ^^^^^^^^^^^
SyntaxError: The requested module 'did-resolver' does not provide an export named 'DIDDocument'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:124:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:190:5)

Node.js v18.17.0
```

Comparing the `ICredentialStatusManager.js` compiled with esbuild and tsc shows some differences:
tsc:
```
export {};
//# sourceMappingURL=ICredentialStatusManager.js.map
```
esbuild:
```
import { DIDDocument, DIDResolutionOptions, DIDResolutionResult } from "did-resolver";
export {
  DIDDocument,
  DIDResolutionOptions,
  DIDResolutionResult
};
//# sourceMappingURL=ICredentialStatusManager.js.map
```

Even when removing it from the `ICredentialStatusManager.ts` file, it will throw similar errors why types or so that are generated locally.

At this point I am not sure if this can be solved by a parameter that has to be passed to esbuild or changed in tsconfig or if we need to update the commands for the following tasks.

## Quality
Check all that apply:
* [ ] I added unit tests.

## Details

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "veramo",
   "private": true,
+  "type": "module",
   "scripts": {
     "build": "cross-env NODE_OPTIONS=\"--experimental-vm-modules --experimental-json-modules\" pnpm build:js && pnpm build:api && pnpm build:schema",
     "build:js": "pnpm -r --stream build",
@@ -88,5 +89,8 @@
   },
   "workspaces": [
     "packages/*"
-  ]
+  ],
+  "dependencies": {
+    "esbuild": "^0.19.12"
+  }
 }
--- a/packages/cli/esbuild.js
+++ b/packages/cli/esbuild.js
@@ -6,7 +6,7 @@ import { exec } from 'child_process';
 build({
     entryPoints: ['src/cli.ts'], // Your entry file
     bundle: false,
-    outfile: 'build/index.cjs.js', // Output file for CommonJS
+    outfile: 'build/cli.cjs.js', // Output file for CommonJS
     format: 'cjs', // Specify CommonJS format
     platform: 'node', // Platform target
     sourcemap: true, // Enable source maps (optional)
--- a/packages/cli/esbuild.js
+++ b/packages/cli/esbuild.js
@@ -1,0 +1,42 @@
+import { build } from 'esbuild';
+import { rmSync, existsSync } from 'fs';
+import { exec } from 'child_process';
+
+// Building for CommonJS
+build({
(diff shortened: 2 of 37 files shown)


mirceanis:
It would be much easier to test the migration of a single package to esbuild and to see if it actually improves things.
I suggested the `@veramo/utils` package as it is relatively isolated, but still uses tricky dependencies.
It's the dependencies that have caused the most issues for folks.

cre8:
Despite the fact that esbuild is fast, I am getting a lot of problems when I don't only want it for compilation, but also being interested in the types etc.
I also tried the approach with compiling it with tcs for esm and in parallel with tsc for commonjs. Problem here is to pack both in one npm package. When you add `"type": "module"` to the package.json, it is treating all files with an '.js' ending as esm, even when they are in the folder for cjs. So according to the error you need to update the file endings for all files that are commonjs from '.js' to '.cjs'. This is not supported by tsc out of the box. I found some projects, but they seem not to be so stable anymore.

Maybe with a bundler like rollup or babel this could work, but I don't have the time right now to dig deeper.

martines3000:
For building both ESM and CommonJS I would recommend [Tsup](https://github.com/egoist/tsup). I'm not sure if it resolves the issues here, but it is a great tool that solves a lot of issues regarding building/bundling for multiple environments. It is also fast as it uses ESBuild under the hood.

PR #1334, opened 2024-01-30
fix(credential-eip712): chainId for any did method

simonas-notcat:

Getting correct chainId for any did method

## Quality
Check all that apply:
* [X] I added unit tests.

### DIFF
--- a/packages/utils/src/did-utils.ts
+++ b/packages/utils/src/did-utils.ts
@@ -162,10 +162,17 @@ export function getEthereumAddress(verificationMethod: VerificationMethod): stri
  * @beta This API may change without a BREAKING CHANGE notice.
  */
 export function getChainId(verificationMethod: _NormalizedVerificationMethod): number {
+  let result
   if (verificationMethod.blockchainAccountId?.includes('@eip155')) {
-    return parseInt(verificationMethod.blockchainAccountId!.split(':').slice(-1)[0])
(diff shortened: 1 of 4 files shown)


packages/utils/src/did-utils.ts
```diff
@@ -161,7 +161,7 @@ export function getEthereumAddress(verificationMethod: VerificationMethod): stri
  *
  * @beta This API may change without a BREAKING CHANGE notice.
  */
-export function getChainIdForDidEthr(verificationMethod: _NormalizedVerificationMethod): number {
+export function getChainId(verificationMethod: _NormalizedVerificationMethod): number {
   if (verificationMethod.blockchainAccountId?.includes('@eip155')) {
```
stanleyyconsensys, line 165: may be we can also add is NAN as the parseInt result 

if it is nan, then throw error, hence the Eip712 fallback to use chainId 1?

stanleyyconsensys (changes requested):
LGTM, leave some comments

packages/utils/src/did-utils.ts
mirceanis: good point

mirceanis (approved):
Looks good.
It is worth handling the NaN case that @stanleyyconsensys mentioned.

The test would be against a malformed blockchainAccountId:
eip155:0xe708:0x197...
eip155:linea:0x197...
0x197...@eip155:0xe708
0x197...@eip155:linea

mirceanis (approved):
Looks good

PR #1365, opened 2024-03-27
fix: address edge case when getting DIDDoc component

### LINKED ISSUE #1364: Unable to Pack DIDComm Message in certain situations
**Bug severity**
3

**Describe the bug**
There is a bug (found while updating the Agent Explorer) where the Veramo agent is unable to correctly pack DIDComm messages when trying to send a message from a `peer-did` and gives a 'could not map an agent key to an skid` error. 

Strangely, this bug only occurs on the 2nd or later `packDIDCommMessage` call of an agent's lifespan. The first call correctly packs the message, but subsequent calls have failures.

**To Reproduce**
You can see this error by running this branch of the Agent Explorer:
https://github.com/veramolabs/agent-explorer/commit/05789564c704c84f82dd054841c178f618ee5465

 - Veramo: v5.6.1-next.57

nickreynolds:

`fixes #1364`

## What is being changed
The `getDIDComponentById` function in `did-resolver/src/resolver.ts` is modified so that it correctly finds the result when the `didUrl` passed in is a keyID without the DID prefixed to it, but the resolved DID Document lists the key with the prefix.

## Quality
Check all that apply:
* [ ] I added unit tests.
* [X] I did not add automated tests because I am unable to reproduce this error in unit tests, and I am aware that a PR without tests will likely get rejected.

## Details
it's possible that this is not the ideal solution and the proper fix for the bug is upstream from this function

### DIFF
--- a/packages/did-resolver/src/__tests__/resolver.test.ts
+++ b/packages/did-resolver/src/__tests__/resolver.test.ts
@@ -1,5 +1,5 @@
 import { DIDResolverPlugin } from '../resolver.js'
-import { Resolver } from 'did-resolver'
+import { Resolver, VerificationMethod } from 'did-resolver'
 
 describe('@veramo/did-resolver', () => {
   it('should throw error when misconfigured', () => {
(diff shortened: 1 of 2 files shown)


mirceanis (approved):
Looks good.
It would be nice to add a test case for this

simonas-notcat (approved):

PR #1411, opened 2024-08-26
PR generated to test if all tests work

veromassera:
## Quality
Check all that apply:
* [ ] I added unit tests.

## Details
PR generated to test if all tests work


nickreynolds:
@veromassera I did some more digging, and it appears that an integration test is failing because a 3rd party service we used (`did.actor`) is no longer operational (which is why the tests started failing even though we didn't merge any changes). I'm looking to find a replacement, or will otherwise just remove the 2 failing tests. I'm not sure it will fix your issue, as I think the failed tests I was seeing on your PR were unrelated and started before this service was turned off, but I will let you know when it's fixed and we can check again

veromassera:
Hi Nick @nickreynolds ,
Thanks for looking into this.
I think my PR should be good to go once the test issues are fixed.
I was running into a problem with a React test dependency, but I managed to sort it out using vm-browserify.

I look forward to hearing from you.

PR #1423, opened 2024-10-28
feat: improve key creation/import options when creating identifiers

0x62:

At the moment, specifying key options when creating identifiers (via `didManagerCreate`) is fragile. Each provider has a slightly different format for `options`, and many omit the ability to provide `meta` for keys. This is a problem, as it means there is no nice way to create both an identifier and associated keys if you need to store metadata.

This PR proposes some non-breaking changes to each of the DID providers that unifies key option signature with a base type. It passes all key options from `options.key` to `keyManagerCreate/keyManagerImport` automatically.

If no `options.key` or `options.keyRef` is provided, each provider falls back to the currently support configuration signature, which has been deprecated.

## What is being changed

`createIdentifier` methods on each DID provider now utilise a base options type which includes a `keyRef` option, allowing the user to pass in an existing key, or a `key` option specifying the details for the key to be created/imported.

```
export type ImportOrCreateKeyOptions<TKey extends TKeyType = TKeyType> = Omit<
  IKeyManagerCreateArgs & MinimalImportableKey,
  'kms' | 'type'
> & { type: TKey }

export type CreateIdentifierBaseOptions<T extends TKeyType = TKeyType> = {
  keyRef?: string;
  key?: ImportOrCreateKeyOptions<T>;
}
```

For `did:peer`, there is an additional `agreementKey` and `agreementKeyRef` option.

Providers

- did-provider-ethr
- did-provider-jwk
- did-provider-key
- did-provider-peer
- did-provider-pkh
- did-provider-web

This change **does not** affect `did-provider-ion`. The ION provider requires specific metadata in each of the keys, which requires some consideration on the best way to handle. 

```
// Create with an existing key
agent.didManagerCreate({
  kms: 'local'.
  provider: 'did:peer',
  options: {
    keyRef: existingKey.kid
  }
})

// Create a new key with params
agent.didManagerCreate({
  kms: 'local'.
  provider: 'did:key',
  options: {
    key: {
      type: 'X25519',
      meta: { userId: 123 }
    }
  }
})

// Import a key
agent.didManagerCreate({
  kms: 'local'.
  provider: 'did:key',
  options: {
    key: {
      type: 'X25519',
      privateKeyHex: 'DEADBEEF',
      meta: { userId: 123 }
    }
  }
})
```

## Quality
Check all that apply:
* [X] I added unit tests.

### DIFF
--- a/packages/did-provider-ethr/package.json
+++ b/packages/did-provider-ethr/package.json
@@ -12,6 +12,7 @@
   "dependencies": {
     "@veramo/core-types": "workspace:^",
     "@veramo/did-manager": "workspace:^",
+    "@veramo/utils": "workspace:^",
     "debug": "^4.3.3",
     "ethers": "^6.11.1",
     "ethr-did": "^3.0.21"
--- a/packages/did-provider-pkh/package.json
+++ b/packages/did-provider-pkh/package.json
@@ -12,6 +12,7 @@
   "dependencies": {
     "@veramo/core-types": "workspace:^",
     "@veramo/did-manager": "workspace:^",
+    "@veramo/utils": "workspace:^",
     "caip": "^1.1.0",
     "debug": "^4.3.3",
     "did-resolver": "^4.1.0",
--- a/packages/did-provider-web/package.json
+++ b/packages/did-provider-web/package.json
@@ -12,6 +12,7 @@
   "dependencies": {
     "@veramo/core-types": "workspace:^",
     "@veramo/did-manager": "workspace:^",
+    "@veramo/utils": "workspace:^",
     "debug": "^4.3.3"
   },
   "devDependencies": {
--- a/packages/did-provider-ethr/src/__tests__/ethr-did-provider.test.ts
+++ b/packages/did-provider-ethr/src/__tests__/ethr-did-provider.test.ts
@@ -65,6 +65,32 @@ describe('EthrDIDProvider', () => {
   })
 
   describe('adding keys', () => {
+    it('supports implicit creation of a Secp256k1 key', async () => {
+      const newDid = await agent.didManagerCreate({
+        kms: KMS,
+        provider: PROVIDER
+      })
+
+      expect(newDid.controllerKeyId).toBeDefined()
+    })
+
+    it('supports creation with a key reference', async () => {
+      const newKey = await agent.keyManagerCreate({
+        kms: KMS,
+        type: 'Secp256k1'
+      })
+
+      const newDid = await agent.didManagerCreate({
+        kms: KMS,
+        provider: PROVIDER,
+        options: {
+          keyRef: newKey.kid,
+        }
+      })
+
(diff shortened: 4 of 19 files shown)


mirceanis (commented):
Yes, there is definitely appetite for including this kind of work. 
Thank you for proposing this!!

--

* try to reduce code duplication, like for importOrGenerateKey
* What do you think about authKey being just key for did:peer? It would allow a simpler options type to be extended by all did providers 
* specifying just the kid would be a simpler API, even when requiring 2 calls to create a DID from a predefined key, but it would be a breaking change.

mirceanis:
> I'd be in favour of keeping the option to do both. As currently implemented, it is non-breaking. I did consider a signature like `key: { kid: string; }`, but this also has ambiguity in whether you're looking up a key by KID or providing a KID for the key being created. `keyRef` has precedent in Veramo (e.g `keyManagerSign({ keyRef: 'xyz' })`), as opposed to a new signature

Agreed. Naming the internal property `kid` was a mistake as it's usually a public JWT header property confusing everyone.
We tried to migrate to `keyRef` at some point but not all the way through. I suppose a simple `id` property would have been better.. or `alias` would have worked as well.

Suggestions welcome!

packages/did-provider-ethr/package.json
```diff
@@ -12,6 +12,7 @@
   "dependencies": {
     "@veramo/core-types": "workspace:^",
     "@veramo/did-manager": "workspace:^",
+    "@veramo/utils": "workspace:^",
```
mirceanis, line 15: I think we should also have a reference in the `did-provider-ethr/tsconfig.json` for this workspace package

packages/did-provider-pkh/package.json
```diff
@@ -12,6 +12,7 @@
   "dependencies": {
     "@veramo/core-types": "workspace:^",
     "@veramo/did-manager": "workspace:^",
+    "@veramo/utils": "workspace:^",
```
mirceanis, line 15: I think we should also have a reference in the `did-provider-pkh/tsconfig.json` for the utils package

packages/did-provider-web/package.json
```diff
@@ -12,6 +12,7 @@
   "dependencies": {
     "@veramo/core-types": "workspace:^",
     "@veramo/did-manager": "workspace:^",
+    "@veramo/utils": "workspace:^",
```
mirceanis, line 15: I think we should also have a reference in the `did-provider-web/tsconfig.json` for this workspace package

mirceanis (changes requested):
Generally looks good.

There are some typos to be fixed for did:peer:2 and I think some workspace dependencies need to be referenced in their corresponding tsconfig files.

0x62:
Thanks for the review @mirceanis, should all be fixed

mirceanis (approved):
Looks great, thank you for the contribution!

PR #1452, opened 2025-04-30
feat(did-manager): add 'isLocal' in 'options' to only modify DIDStore

arwykchorres:
## What is being changed

I've added `isLocal` to allow only modifying DIDStore locally for the next methods, this allows modifying DIDStore without calling any DID provider.

- didManagerAddKey
- didManagerRemoveKey
- didManagerAddService
- didManagerRemoveService

## Motivations

I have a use case where an Agent gives access to the DID to another Agent so that this one can also manage the DID. Actually doing that is only possible if I modify the database directly so the change I make allows the process to be done by the same methods.

## Quality
Check all that apply:
* [ ] I added unit tests.


packages/did-manager/src/id-manager.ts
```diff
@@ -223,7 +223,11 @@ export class DIDManager implements IAgentPlugin {
   ): Promise<any> {
     const identifier = await this.store.getDID({ did })
     const provider = this.getProvider(identifier.provider)
-    const result = await provider.addKey({ identifier, key, options }, context)
+    let result = true
+    // @ts-ignore
```
mirceanis, line 227: I think it would be best to define options in the argument types `IDIDManagerAddKeyArgs` and so on instead of ts-ignore.

The IDE should show you the typescript warnings. Otherwise, run `pnpm build` locally to see that everything builds.

The PR template has instructions for this in one of the check boxes

PR #1501, opened 2026-07-29
fix(did-comm): add required `apv` protected header when packing JWE messages

### LINKED ISSUE #1500: `packDIDCommMessage` omits the required `apv` protected header (anoncrypt and authcrypt)
### Describe the bug

`packDIDCommMessage()` never computes `apv`. The JWE protected header of both
anoncrypt and authcrypt envelopes contains only `alg`, `typ`, `epk`, `enc`
(plus `skid` for authcrypt). Strict DIDComm v2 implementations treat `apv` as
a required protected-header field and reject the envelope during parsing,
before any decryption is attempted.

### What the spec says

DIDComm v2, [ECDH-ES section](https://identity.foundation/didcomm-messaging/spec/v2.0/#ecdh-es-key-wrapping-and-common-protected-headers)
(same definition in the ECDH-1PU section):

> `apv`: this represents the recipients' `kid` list. The list must be
> alphanumerically sorted, `kid` values will then be concatenated with a `.`
> and the final result MUST be base64 URL (no padding) encoding of the SHA256
> hash of concatenated list.

Reference implementations agree: in didcomm-rust the field is non-optional —
`ProtectedHeader.apv` is a plain `&str` (not `Option`) in
[`src/jwe/envelope.rs`](https://github.com/sicpa-dlab/didcomm-rust/blob/main/src/jwe/envelope.rs),
and `ParsedJWE::verify_didcomm()` in
[`src/jwe/parse.rs`](https://github.com/sicpa-dlab/didcomm-rust/blob/main/src/jwe/parse.rs)
recomputes the hash from the recipient kids and rejects the message on
mismatch ("APV mismatch").

### Observed behavior

Sending an anoncrypt message (`ECDH-ES+A256KW`, tested with both `A256GCM`
and `XC20P`) packed by `packDIDCommMessage()` to a didcomm-rust-based
receiver fails with HTTP 400:

```
DIDCommMalformed: Malformed: Unable parse protected header: missing field `apv` at line 1 column 166
```

Re-packing the same message with `apv = base64url(sha256(sortedKids.join('.')))`
added to the protected header *and* passed into the encrypters is accepted
(HTTP 202), and the resulting envelope also passes didcomm-rust's
`verify_didcomm()` check locally.

### Where in the code

- `packages/did-comm/src/didcomm.ts` — `packDIDCommMessageJWE()` has the full
  recipient list before creating encrypters, but never computes `apv`; the
  encrypter factory calls (e.g. `xc20pAnonEncrypterX25519WithA256KW(recipient.publicKeyBytes, recipient.kid)`)
  omit the optional `apv` argument.
- The low-level layer is already prepared: every anon encrypter in
  `packages/did-comm/src/encryption/a256kw-encrypters.ts` and
  `xc20pkw-encrypters.ts` accepts `apv?: string` and forwards it to the
  Concat KDF via `createFullEncrypter()`.

### Suggested fix

After collecting recipients in `packDIDCommMessageJWE()`:

```ts
const kids = recipients.map((r) => r.kid).sort()
const apv = bytesToBase64url(sha256(stringToUtf8Bytes(kids.join('.'))))
protectedHeader = { ...protectedHeader, apv }
// and pass `apv` to each encrypter so it participates in the KDF
```

Backwards compatibility: did-jwt ≥ 7.x merges the protected header into each
recipient header before KEK computation on decrypt (`decryptJWE`), so older
Veramo receivers can still decrypt envelopes that carry `apv`.

onyxblade:
Fixes #1500

## Problem

`packDIDCommMessageJWE()` never computes `apv`. The JWE protected header of both anoncrypt and authcrypt envelopes contains only `alg`, `typ`, `epk`, `enc` (plus `skid` for authcrypt). Strict DIDComm v2 implementations treat `apv` as a required protected-header field and reject the envelope during parsing, before any decryption is attempted — e.g. didcomm-rust fails with:

```
DIDCommMalformed: Malformed: Unable parse protected header: missing field `apv` at line 1 column 166
```

The [DIDComm v2 spec](https://identity.foundation/didcomm-messaging/spec/v2.0/#ecdh-es-key-wrapping-and-common-protected-headers) defines it identically for ECDH-ES and ECDH-1PU:

> `apv`: this represents the recipients' `kid` list. The list must be alphanumerically sorted, `kid` values will then be concatenated with a `.` and the final result MUST be base64 URL (no padding) encoding of the SHA256 hash of concatenated list.

## Changes

- New `computeApv(kids)` helper in `packages/did-comm/src/utils.ts` implementing the spec definition.
- `packDIDCommMessageJWE()` computes `apv` over all recipients (including `bcc`) once the recipient list is known, adds it to the protected header, and passes it to **every** encrypter so it also participates in the Concat KDF as PartyVInfo.

The low-level layer needed no changes — every encrypter in `a256kw-encrypters.ts` / `xc20pkw-encrypters.ts` already accepted an optional `apv` and forwarded it through `createFullEncrypter()`. Only the 12 call sites were passing it.

## Compatibility

`decryptJWE()` in did-jwt >= 7 merges the protected header into each recipient header before computing the KEK, so receivers on current did-jwt derive the same key and can still decrypt envelopes that carry `apv`. All existing pack/unpack round-trip tests pass unchanged.

## Tests

Added an `apv protected header` suite to `packing.test.ts`:

- `computeApv` checked against an independent `node:crypto` implementation, including sort-order independence and the no-padding base64url alphabet.
- A 12-case matrix over {anoncrypt, authcrypt} × {XC20P, A256GCM, A256CBC-HS512} × both key-wrap algs, asserting the protected header `apv` matches the hash of the actual recipient `kid`s — one case per encrypter call site touched.
- Multi-recipient and `bcc` coverage, plus recipient-order independence.

Note that a successful round trip is itself an assertion that `apv` reached the encrypters: since the protected header is merged into every recipient header on decrypt, an `apv` that did not participate in the KDF at pack time yields a different KEK. Dropping `apv` from a single encrypter call locally fails 5 tests, confirming the matrix is not vacuous.

`pnpm jest packages/did-comm` — 170 passed. `__tests__/localAgent.test.ts` — 219 passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/did-comm/src/__tests__/packing.test.ts
+++ b/packages/did-comm/src/__tests__/packing.test.ts
@@ -9,6 +9,8 @@ import { DIDResolverPlugin } from '../../../did-resolver/src'
 import { type DIDDocument, Resolver } from 'did-resolver'
 import { type IDIDComm } from '../types/IDIDComm.js'
 import { base64ToBytes, bytesToUtf8String } from '@veramo/utils'
+import { createHash } from 'node:crypto'
+import { computeApv } from '../utils.js'
 
 const multiBaseDoc = {
   '@context': ['https://www.w3.org/ns/did/v1', 'https://w3id.org/security/suites/jws-2020/v1'],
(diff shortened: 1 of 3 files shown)


mirceanis (approved):
looks good. Thank you for the contribution

